PR #9, opened 2016-08-07
Client certificate authentication support

silasbw:


lib/api.js
```diff
@@ -17,6 +17,17 @@ class Api {
     this.url = options.url;
     this.version = options.version || 'v1';
     this.path = `/api/${ this.version }`;
+    this.requestOptions = options.request || {};
```
jcrugzz: We should document this as well.

jcrugzz:
LGTM 👍. Just minor thing on docs

PR #13, opened 2016-08-15
Increase `test-integration` timeout to 30 seconds

silasbw:
10 seconds is too short and the tests were timing out


jcrugzz:
LGTM 👍

PR #14, opened 2016-08-15
Cluster level resources, for example, `k8.resourcequotas.get`

silasbw:
This removes the default namespace shortcuts (e.g., `api.rc`) because they're
ambiguous with the cluster level resources.


jcrugzz:
LGTM 👍

PR #18, opened 2016-08-31
Throw an Error if `rc.po.get` invoked without options and cb.

silasbw:
`rc.po.get` breaks usual pattern (if no options, get everything), because it's
not useful do that (users should use `po.get`), and we don't handle it in the
`rc.po.get` code. It's a programming mistake of `options` is missing, so throw
an Error.

### DIFF
--- a/lib/replicationcontrollers.js
+++ b/lib/replicationcontrollers.js
@@ -29,6 +29,10 @@ class ReplicationControllerPods extends BaseObject {
    * @param {callback} cb - The callback that handles the response
    */
   get(options, cb) {
+    if (arguments.length < 2) {
+      throw new Error(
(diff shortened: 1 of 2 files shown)


lib/replicationcontrollers.js
```diff
@@ -29,6 +29,10 @@ class ReplicationControllerPods extends BaseObject {
    * @param {callback} cb - The callback that handles the response
    */
   get(options, cb) {
+    if (arguments.length < 2) {
+      throw new Error(
+        'GETing ReplicationController Pods requires options and cb arguments');
```
bmeck: dunno if you want to add a note that they might want to use po.get directly

bmeck:
seems reasonable, LGTM

PR #20, opened 2016-09-10
lookup api path by version

kaija:
lookup api path for version prefix start with autoscaling / batch and extensions.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -16,7 +16,7 @@ class Api {
   constructor(options) {
     this.url = options.url;
     this.version = options.version || 'v1';
-    this.path = `/api/${ this.version }`;
+    this.path = '/' + lookupApi(options.version) + '/' + this.version;
     this.requestOptions = options.request || {};
 
     this.requestOptions.ca = options.ca;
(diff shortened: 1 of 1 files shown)


silasbw:
Hey, thanks for submitting this.  It's clear we need to support API groups better.  I think to do this in a clean and extensible way we'll probably want an approach like the official Kubernetes Go client.  E.g., https://github.com/kubernetes/client-go/blob/master/1.4/kubernetes/clientset.go#L42

One way to do this with `kubernetes-client` is to make [`Api`](https://github.com/godaddy/kubernetes-client/blob/master/lib/api.js) more general purpose.  Then `kubernetes-client` would implement something like:

```js
class Core extends Api {
  // The core /api/version
}

class ApisExtension extends Api {
  // /apis/extensions/version
}
```

ernoaapa:
:+1: Needed this also because otherwise cannot access deployments. Would be cool one way or another to get merged. Temporarily I will use @kaija's branch.

TWood67:
@ernoaapa @kaija @silasbw I am working on a permanent solution and will have a PR up tonight.

silasbw:
Thanks @TWood67 !

silasbw:
@TWood67 submitted https://github.com/godaddy/kubernetes-client/pull/23.  I think it's a good long term direction, and it seems like we're going to merge it soon.  We'll push a new version to npm.  Thanks again @kaija.

PR #23, opened 2016-09-14
Apis extension

TWood67:
Created a more permanent solution for the k8s apis extension path.


lib/api.js
```diff
@@ -13,10 +13,10 @@ class Api {
    * @param {string} options.version - Kubernetes API version
    * @param {string} options.namespace - Default namespace
    */
-  constructor(options) {
+  constructor(options, genericTypes) {
```
silasbw: Let's fold `genericTypes` into the `options` object to be consistent with the other arguments we're passing to this constructor (they're all passed via the `options` object)

lib/apisExtension.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var Api = require('./api');
+
+class ApisExtension extends Api {
+  constructor(options) {
+    options.path = 'apis/extensions';
```
silasbw: Instead of mutating `options`, can we use `Object.assign` to make a copy?  e.g.,
```
options = Object.assign({}, options, {
  version: options.version || 'v1beta1',
  genericTypes: genericTypes
});
```

lib/core.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+var Api = require('./api');
+var aliasResources = require('./common').aliasResources;
+
+class Core extends Api {
+  constructor(options) {
+    options.path = 'api';
```
silasbw: Same comment as above about using `Object.assign`

lib/index.js
```diff
@@ -1,4 +1,5 @@
-module.exports = require('./api');
+module.exports.Core = require('./core');
```
silasbw: should we keep a `module.exports = require('./core');` to be backwards compatible?  @jcrugzz what do you think?

lib/apisExtension.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
silasbw: Filename nit: let's name this file `extensions.js`

test/api.test.js
```diff
@@ -1,3 +1,4 @@
+/*
```
silasbw: remove?

test/extensions.test.js
```diff
@@ -0,0 +1,207 @@
...
+              name: 'test',
+              imagePullPolicy: 'Never',
+              image: 'test.com:5000/test:v0'
+            }]
+          }
+        }
+      }
+    };
+
+    beforeTesting('int', done => {
```
silasbw: you can do:
```
beforeTesting('int', api.wipe)
```
mocha will fail if the cb is passed an error

test/extensions.test.js
```diff
@@ -0,0 +1,207 @@
...
+      ], (err, results) => {
+        assume(err).is.falsy();
+        const deployments = results[0];
+        assume(deployments.metadata.name).is.equal(resourceName);
+        done();
+      });
+    });
+
+    describe('lists', () => {
+      beforeTesting('int', done => {
```
silasbw: same

test/extensions.test.js
```diff
@@ -0,0 +1,207 @@
...
+                imagePullPolicy: 'Never',
+                image: 'test.com:5000/test:v0'
+              }
+            ]
+          }
+        }
+      }
+    };
+
+    beforeTesting('int', done => {
```
silasbw: same

test/extensions.test.js
```diff
@@ -0,0 +1,207 @@
...
+        assume(err).is.falsy();
+        const ds = results[0];
+        assume(ds.metadata.name).is.equal(resourceName);
+        done();
+      });
+    });
+
+    describe('lists', () => {
+      beforeTesting('int', done => {
+        api.wipe(err => {
```
silasbw: same

PR #28, opened 2016-09-25
Rename `Api` to `ApiGroup`

silasbw:
"Group" is terminology from: https://github.com/kubernetes/kubernetes/blob/master/docs/api.md


jcrugzz:
👍 On rename

PR #29, opened 2016-09-26
Ensure API for Kubernetes objects is consistent.

silasbw:
Before this change our constructors for ReplicationControllers, Pods, and
Namespaces had different APIs than the BaseObject. This fixes the mismatch.


jcrugzz:
This seems reasonable to me 👍

PR #30, opened 2016-09-27
Add a wrapper for switching API clients

silasbw:
E.g.,
```
const api = new Api(...);
api.core.ns.po.get(...);
api.group('extensions/v1beta1').ns.deployments.get(...);
api.group(manifest).ns.kind(manifest).post(...);
```

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -28,17 +28,17 @@ class Api {
   group(v) {
     const groupVersion = v.apiVersion || v;
     const pieces = groupVersion.split('/');
-    let group;
+    let Group;
     let version;
     if (pieces.length > 1) {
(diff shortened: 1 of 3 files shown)


lib/api.js
```diff
@@ -0,0 +1,45 @@
...
+    let version;
+    if (pieces.length > 1) {
+      group = groups[pieces[0].toLowerCase()];
+      version = pieces[1];
+    } else {
+      group = Core;
+      version = pieces[0];
+    }
+    const options = Object.assign({}, this.options,  { version });
+    return new group(options);
```
jcrugzz: capitalize `group` since it is a constructor

PR #35, opened 2016-10-25
Fixed an issue with aliasing and `kind` error

TWood67:
The `kind` method was not bound correctly and would throw an error stating that `kind` was not a function. I also noticed an issues where resources that ended in `es` were not being aliased to their singular version correctly. An example is `ingresses` would alias to `ingresse`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ correct Kubernetes API group and version to use based on manifests:
 const K8Api = require('kubernetes-client');
 const api = new K8Api.Api({
   url: 'http://my-k8-api-server.com',
-  namespace: 'test-ns'
+  namespace: 'my-project'
 });
 
 const manifest0 = {
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ correct Kubernetes API group and version to use based on manifests:
 const K8Api = require('kubernetes-client');
 const api = new K8Api.Api({
   url: 'http://my-k8-api-server.com',
+  namespace: 'test-ns'
 });
 
 const manifest0 = {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -80,6 +80,7 @@ correct Kubernetes API group and version to use based on manifests:
 const K8Api = require('kubernetes-client');
 const api = new K8Api.Api({
   url: 'http://my-k8-api-server.com',
+  namespace: 'test-ns'
```
silasbw: we use `namespace: 'my-project'` in the example above.  Can you use `'my-project'` here too?  or change other uses to `'test-ns'`?

lib/common.js
```diff
@@ -25,13 +25,23 @@ module.exports.aliasResources = function (resourceObject) {
     services: ['svc']
   };
 
+  const esRemoval = [
```
silasbw: Rename `esRemoval` to `pluralsWithEs` or `esPlurals`

Also, make it an object.  this shorthand works in node.js >= 4.X
```
{
  componentstatuses,
  ingresses
}
```

lib/common.js
```diff
@@ -25,13 +25,23 @@ module.exports.aliasResources = function (resourceObject) {
...
+
   for (const resourceType of Object.keys(resourceAliases)) {
     if (resourceObject[resourceType]) {
       for (const alias of resourceAliases[resourceType]) {
         resourceObject[alias] = resourceObject[resourceType];
       }
-      // Alias without the ending 's'
-      const single = resourceType.substr(0,  resourceType.length - 1);
+
+      let length = 1;
```
silasbw: `const trimLength = esPlurals[resourceType] ? 2 : 1;`

jcrugzz:
LGTM 👍

PR #39, opened 2016-11-24
Add a name for PersistentVolumes/PVs

ankon:
Without this change I was unable to create/modify persistent volumes.


silasbw:
+1, thanks

silasbw:
@ankon I published your fix as version 2.2.3 to npmjs

PR #40, opened 2016-12-06
Add namespaced support to objects other than `namespaces`

silasbw:
There are several objects/entities that are namespaced by Kubernetes objects.
`/api/v1/namespaces/{namespace}/pods/{name}/log` is an obvious one, but there's
more. For example:
* Proxy objects (e.g., `/api/v1/namespaces/{namespace}/services/{name}/proxy/`)
* Deployment related objects (e.g., `/apis/extensions/v1beta1/namespaces/{namespace}/deployments/{name}/rollback`)
* ...

This PR implements a story for handling these cases and changes the
documentation to encourage users to use namespaced functionality by default
(e.g., `api.ns.po('foo').get(print)` vs `api.ns.po.get(foo, print)`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ function print(err, result) {
   console.log(JSON.stringify(err || result, null, 2));
 }
 
-k8.namespaces.replicationcontrollers.get('http-rc', print);
+k8.namespaces.replicationcontrollers('http-rc').get(print);
 ```
 
 kubernetes-client supports the Extensions API group. For example, GET
@@ -49,7 +49,7 @@ const k8Ext = new K8Api.Extensions({
   namespace: 'my-project' // Defaults to 'default'
 });
 
-k8Ext.namespaces.deployments.get('http-deployment', print);
+k8Ext.namespaces.deployments('http-deployment').get(print);
 ```
 
 ### Creating and updating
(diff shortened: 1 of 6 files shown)


silasbw:
This is in an RFC state and isn't complete (e.g., missing tests).  If folks like the API changes and direction I'll finish this off.

jcrugzz:
@silasbw I love what you are doing here and that it's so few lines of code. I would add more comments explaining what you are doing in the base class for future contributor understanding and continue along this thread with tests etc. 

This will be a breaking change but I'm ok with that.

bmeck:
@silasbw love the change, would love to get the template engine change in at same time as this if possible to minimize break but also make the bump a bit bigger.

silasbw:
@jcrugzz actually, the old way (e.g., `api.ns.po.get('foo', print)`) still works, it's just not documented.

@bmeck what's the template engine -- are you thinking of the one associated with `trimmer`?

I'll proceed with this bad larry, add tests, comments, etc.  Thanks for the feedback.

TWood67:
@silasbw I think this is a great direction to go. Good stuff!

silasbw:
I took a pass at fleshing this PR out.  I think it's ready for a review.

jcrugzz:
THis looks good 👍

PR #48, opened 2016-12-19
Improve integration testing framework to give more consistent results

silasbw:
Each time we need a fresh namespace, generate a new name, instead of recycling
(deleting and re-creating) a single name. Removing a namespace isn't
synchronous and can be stuck in 'Terminating' for a while, especially on
minikube. That resulted in timeouts, or very long running tests.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -1,7 +1,8 @@
-/* eslint no-process-env:0 */
+/* eslint no-process-env:0 no-sync:0 */
 'use strict';
 
 const async = require('async');
+const crypto = require('crypto');
 const fs = require('fs');
 const path = require('path');
 const yaml = require('js-yaml');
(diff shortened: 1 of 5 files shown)


test/common.js
```diff
@@ -78,78 +81,109 @@ if (testing('int')) {
...
+
+if (!testing('int')) {
+  changeNameUnit();
+}
+
+//
+// Do our best to cleanup namespaces, but it's hard to ensure we don't leak.
+//
+function atexit() {
+  if (testing('int') && module.exports.currentName) {
```
jcrugzz: yea this probably wont work but not the biggest deal. Maybe we should have an `after` that gets run that clears all namespaces when we are `testing('int')`? I think thats the cleanest way to try and do this for real.

jcrugzz:
LGTM overall but would be nice to know that our instances have had a "delete-all-namespaces" command run at the end.

Issue #44, opened 2016-12-13
In-cluster config support


hekike:
Hi,

would you be open for adding a built-in in-cluster config?

Based on:
https://github.com/kubernetes/client-go/blob/124670e99da15091e13916f0ad4b2b2df2a39cd5/rest/config.go#L274
and
http://kubernetes.io/docs/user-guide/accessing-the-cluster/#accessing-the-api-from-a-pod

Something like:
```js
const fs = require('fs')

const certPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt'
const tokenPath = '/var/run/secrets/kubernetes.io/serviceaccount/token'
const namespacePath = '/var/run/secrets/kubernetes.io/serviceaccount/namespace'

function getInClusterConfig () {
  const host = process.env.KUBERNETES_SERVICE_HOST
  const port = process.env.KUBERNETES_SERVICE_PORT
  let cert
  let namespace
  let bearer

  if (!host || !port) {
    throw TypeError('Unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined')
  }

  try {
    cert = fs.readFileSync(certPath, 'utf8')
  } catch (err) {
    throw TypeError('Expected to load root CA config from ${certPath}, but got err', err)
  }

  try {
    bearer = fs.readFileSync(tokenPath, 'utf8')
  } catch (err) {
    throw TypeError('Expected to load token config from ${tokenPath}, but got err', err)
  }

  try {
    namespace = fs.readFileSync(namespacePath, 'utf8')
  } catch (err) {
    throw TypeError('Expected to load namespace config from ${namespacePath}, but got err', err)
  }

  return {
    url: `https://${host}:${port}`,
    cert,
    auth: { bearer },
    namespace
  }
}
````

The interface could be:
```js
const K8Api = require('kubernetes-client');
const k8 = new K8Api.Core(K8Api.getInClusterConfig())
```

I'm happy to send a PR if you like the idea.
What do you think?

silasbw:
Yeah, I like this idea.  I PR would be much appreciated.

@jcrugzz any thoughts?

jcrugzz:
Yea this seems super convenient 👍. Would definitely take a PR

PR #49, opened 2016-12-19
feat(core): add in-cluster config helper

hekike:
- [x] add in-cluster config helper

Related to issue #44.
I used JSDoc for code documentation, let me know if you would like to see it in different way.


lib/core.js
```diff
@@ -1,8 +1,58 @@
...
+    const port = process.env.KUBERNETES_SERVICE_PORT;
+    let cert;
+    let namespace;
+    let bearer;
+
+    if (!host || !port) {
+      throw TypeError('Unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined');
+    }
+
+    try {
```
silasbw: I'd simplify this by not handling each `readFileSync` exception individually because I think there's enough context in those `Error`s:
```
const cert = fs.readFileSync(certPath, 'utf8');
const bearer = fs.readFileSync(tokenPath, 'utf8');
const namespace = fs.readFileSync(namespacePath, 'utf8');
```

test/client.test.js
```diff
@@ -0,0 +1,52 @@
...
+      fsReadFileSync
+        .withArgs('/var/run/secrets/kubernetes.io/serviceaccount/token')
+        .returns('my-token')
+
+      fsReadFileSync
+        .withArgs('/var/run/secrets/kubernetes.io/serviceaccount/namespace')
+        .returns('my-namespace')
+
+      const config = K8Api.getInClusterConfig();
+      assume(config).eqls({
```
silasbw: nit: the other few tests that use equals use the full spelling `equals`.  can we use `equals` here to be consistent?

test/client.test.js
```diff
@@ -0,0 +1,52 @@
...
+    sandbox = sinon.sandbox.create();
+  })
+
+  afterEach(() => {
+    sandbox.restore();
+  })
+
+  describe('getInClusterConfig', () => {
+    it('should return with in-cluster config', () => {
+      // may doesn't exist (cannot stub undefined)
```
silasbw: `// You cannot stub undefined, so ensure KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT are defined`

test/client.test.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
```
silasbw: rename this file to `config.test.js`

test/client.test.js
hekike: `equals` is for reference based comprasion: https://www.npmjs.com/package/assume#equal-equals-eq-eqs-exactly
`eqls`: https://www.npmjs.com/package/assume#eql-eqls checks deeply

hekike:
@silasbw I fixed/commented the outstanding issues.

lib/index.js
```diff
@@ -1,8 +1,11 @@
 const core = require('./core');
+const config = require('./config')
+
 module.exports = core;
 module.exports.Api = require('./api');
 module.exports.Core = core;
 module.exports.Extensions = require('./extensions');
+module.exports.getInClusterConfig = config.getInCluster
```
silasbw: `module.exports.config = require(./config);`

I'm anticipating we'll add more config-like options, so it would be good to start out with them namespaced under `.config`.

jcrugzz: yea i like the idea of it being `k8.config.inCluster` personally.

hekike:
I moved it under config scope.

PR #51, opened 2016-12-20
Note in README.md about the potential for disconnects when watching

silasbw:


herb-kuta-lge:
This note was dropped by https://github.com/godaddy/kubernetes-client/pull/222. Was that intentional? The description of `--min-request-timeout` on https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/ still says: 

> An optional field indicating the minimum number of seconds a handler must keep a request open before timing it out. Currently only honored by the watch request handler, which picks a randomized value above this number as the connection timeout, to spread out load.

herb-kuta-lge:
I've confirmed that this is still an issue. @silasbw < Sould I create an issue to add back the note?

PR #54, opened 2016-12-23
fix(aliases): Add missing object aliases

silasbw:
* configmaps (aka 'cm')
* deployments (aka 'deploy')
* endpoints (aka 'ep')


jcrugzz:
LGTM 👍

Issue #53, opened 2016-12-23
Missing support for some object types


silasbw:
clusters, petsets, quota (not resourcequotas/'quota') and replicasets.

ankon:
For Petset have a look at https://github.com/Collaborne/kubernetes-client/tree/collaborne. This name is deprecated though, so I'm currently preparing a PR to support StatefulSet instead: https://github.com/Collaborne/kubernetes-client/tree/collaborne-3.1.0

The other types can probably get added in a similar way.

silasbw:
Awesome!  Looking forward to your PR.

ankon:
@silasbw out of interest: do you have documentation pointers to the "cluster" and "quote" resources?

silasbw:
I got those names out of an older version of [1] and meant to follow up on documentation.  I suspected "quota" was a bugo and it looks like it was removed [2].  I think it's safe to ignore my comment about the distinction between "quota" and "resourcequotas/quota".

[1]: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cmd.go#L172
[2]: https://github.com/kubernetes/kubernetes/commit/ffcc2f61b55a166b21d3e4af3d4c6d97696cf1b6

PR #57, opened 2017-01-11
Support more types

ankon:
Relates to #53

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -3,10 +3,12 @@
 const Core = require('./core');
 const Extensions = require('./extensions');
 const Apps = require('./apps');
+const Batch = require('./batch');
 
 const groups = {
   extensions: Extensions,
-  apps: Apps
+  apps: Apps,
+  batch: Batch
 };
 
 class Api {
@@ -16,12 +18,14 @@ class Api {
    * @param {object} options.core - Optional default Core client
    * @param {object} options.extensions - Optional default Extensions client
    * @param {object} options.apps - Optional default Apps client
+   * @param {object} options.batch - Optional default Batch client
    */
   constructor(options) {
     this.options = options;
     this.core = options.core || new Core(options);
     this.extensions = options.extensions || new Extensions(options);
     this.apps = options.apps || new Apps(options);
+    this.batch = options.batch || new Batch(options);
   }
 
   /**
--- a/lib/api.js
+++ b/lib/api.js
@@ -4,11 +4,13 @@ const Core = require('./core');
 const Extensions = require('./extensions');
 const Apps = require('./apps');
 const Batch = require('./batch');
+const Rbac = require('./rbac');
 
 const groups = {
   extensions: Extensions,
   apps: Apps,
-  batch: Batch
+  batch: Batch,
+  'rbac.authorization.k8s.io': Rbac
 };
 
 class Api {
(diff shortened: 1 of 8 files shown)


lib/apps.js
```diff
@@ -0,0 +1,19 @@
...
+const ApiGroup = require('./api-group');
+
+class Apps extends ApiGroup {
+  constructor(options) {
+    const genericTypes = [
+      'petsets'
+    ];
+    options = Object.assign({}, options, {
+      path: 'apis/apps',
+      version: options.version || 'v1beta1',
```
silasbw: How should we handle the petset -> statefulset rename [1]?

The simplest thing would be to add `statefulset` to the `genericTypes` array and add a comment about the rename.  Any better ideas?

[1]: https://github.com/kubernetes/kubernetes/pull/35663

lib/batch.js
```diff
@@ -0,0 +1,19 @@
...
+const ApiGroup = require('./api-group');
+
+class Batch extends ApiGroup {
+  constructor(options) {
+    const genericTypes = [
+      'scheduledjobs'
+    ];
+    options = Object.assign({}, options, {
+      path: 'apis/batch',
+      version: options.version || 'v2alpha1',
```
silasbw: there's a `v1` batch -- should we make that the default?

lib/batch.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const ApiGroup = require('./api-group');
+
+class Batch extends ApiGroup {
+  constructor(options) {
+    const genericTypes = [
+      'scheduledjobs'
```
silasbw: Similar rename issue as above -- scheduledjobs -> cronjobs [1]

[1]: https://github.com/kubernetes/kubernetes/pull/36021

lib/common.js
```diff
@@ -5,8 +5,11 @@ module.exports.aliasResources = function (resourceObject) {
   //   http://kubernetes.io/docs/user-guide/kubectl-overview/
   // and anything else we think is useful.
   const resourceAliases = {
+    clusterroles: [],
```
silasbw: should we include petsets and scheduledjobs?

lib/namespaces.js
```diff
@@ -37,7 +37,10 @@ class Namespaces extends BaseObject {
     // Generic objects we don't implement special functionality for.
     //
     const genericTypes = [
+      'clusterroles',
```
silasbw: should we include petsets and scheduledjobs?

lib/api.js
```diff
@@ -2,9 +2,15 @@
 
 const Core = require('./core');
 const Extensions = require('./extensions');
+const Apps = require('./apps');
+const Batch = require('./batch');
+const RBAC = require('./rbac');
```
silasbw: Our convention has been to avoid consecutive caps (e.g., `apiUrl`, not `apiURL`).  So this would be `Rbac` not `RBAC`.  `Rbac` looks funny, I know, but it follows the convention we're currently using.

lib/batch.js
ankon: Really good question, I was thinking about that as well. 

I guess it makes sense to have the "conservative" default here, anyone that wants alpha/beta stuff should do so knowingly.

ankon:
I've fixed the nits above and pushed for a quick look at it - but tests are still missing and coming.

This is now rebased this onto #59, while I'm working on the tests :)

silasbw:
Did you make progress on tests?  or get waylaid by that cleanup that needed to happen :)

Is there something I can do to help you finish this?

silasbw:
Rebased/updated branch here: https://github.com/godaddy/kubernetes-client/pull/78

Thanks again @ankon for initiating these changes.

PR #59, opened 2017-01-12
Reduce duplicated code in tests

ankon:
While trying to write tests for #57 I stumbled over a lot of duplication. 

This PR tries to reduce this duplication. 

One area that I still have to understand is the duplication inside the tests -- from what I can see we should be able to write generic tests for each of the types that are marked as 'generic' in the actual library, with the only variations being the actual resource and the API group to use.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -7,8 +7,6 @@ const fs = require('fs');
 const path = require('path');
 const yaml = require('js-yaml');
 
-const Core = require('../lib/core');
-const Extensions = require('../lib/extensions');
 const Api = require('../lib/api');
 
 const defaultName = process.env.NAMESPACE || 'integration-tests';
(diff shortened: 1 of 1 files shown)


test/common.js
```diff
@@ -61,6 +59,23 @@ function newName() {
...
 
+function injectApis(moduleExports, options, versionOverride) {
+  // APIs and their default version
+  const apis = {
+    'api': [ require('../lib/core'), 'v1' ],
+    'extensions': [ require('../lib/extensions'), 'v1beta1' ]
+  }
+  Object.keys(apis).forEach(apiName => {
+    const api = apis[apiName];
+    moduleExports[apiName] = new (Function.prototype.bind.call(api[0], null, Object.assign({}, options, {
```
silasbw: isn't this sufficient in this case:

```
    moduleExports[apiName] = new api[0](Object.assign({}, options, {
      version: versionOverride || api[1],
    }));
```

test/common.js
```diff
@@ -61,6 +59,23 @@ function newName() {
   return `${ defaultName }-${ buffer.toString('hex') }`;
 }
 
+function injectApis(moduleExports, options, versionOverride) {
+  // APIs and their default version
+  const apis = {
+    'api': [ require('../lib/core'), 'v1' ],
```
silasbw: nits: drop the `'`s on the property names, no space after `[` and before `]`.

test/common.js
```diff
@@ -147,25 +144,14 @@ function changeNameInt(cb) {
...
-    version: process.env.VERSION || 'v1beta1',
-    namespace: currentName
-  });
-
-  module.exports.apiGroup = new Api({
-    url: 'http://mock.kube.api',
-    namespace: currentName
-  });
+  injectApis(module.exports, {
+      url: mockUrl,
```
silasbw: nit: it looks like you have extra spaces here

silasbw:
Hey @ankon are you planning on finishing this off?

silasbw:
I rebased master onto this branch and have an updated PR https://github.com/godaddy/kubernetes-client/pull/77.  Continuing this discussion on that PR.  Thank you @ankon for kicking this off.

PR #61, opened 2017-01-20
feat(insecureSkipTlsVerify): plumb an option to skip verifying server certs

silasbw:


lib/api-group.js
```diff
@@ -23,6 +28,10 @@ class ApiGroup {
     this.requestOptions.cert = options.cert;
     this.requestOptions.key = options.key;
 
+    if (options.hasOwnProperty('insecureSkipTlsVerify')) {
+      this.requestOptions.strictSSL = !options.insecureSkipTlsVerify;
```
TWood67: Can we utilize `options.request` for this? Then when constructing the object we use:
```
const k8 = new K8Api({
   url: 'https://my-k8-api-server.com',
   request: {
    strictSSL: false
   },
   auth: {
     user: 'user',
     pass: 'pass'
   }
 });
```

lib/api-group.js
```diff
@@ -23,6 +28,10 @@ class ApiGroup {
     this.requestOptions.cert = options.cert;
     this.requestOptions.key = options.key;
 
+    if (options.hasOwnProperty('insecureSkipTlsVerify')) {
```
bmeck: why hasOwnProperty. if we want that we should also not use it off the object since people could override it

lib/api-group.js
silasbw: Yeah we could.  I think it's useful to pull this into the kubernetes-client API (instead of having it as part of the hacky, pass things to `request` API).  For example `insecureSkipTlsVerify` matches the official Go client option (and the `kubectl`) one; and this gives us more flexibility to move off of `request` if we ever want to.

lib/api-group.js
indexzero: @bmeck is technically correct here (see: [`hasOwnProperty` vs `in`](http://stackoverflow.com/questions/13632999/if-key-in-object-or-ifobject-hasownpropertykey). I believe that the correct usage here would be:

``` js
if ('insecureSkipTlsVerify' in options) {
```

Issue #43, opened 2016-12-11
ThirdPartyResource support


silasbw:
kubernetes-client doesn't support [ThirdPartyResource](http://kubernetes.io/docs/user-guide/thirdpartyresources/).

We should add support to the ApiGroup and add an example usage to README.md.

PR #67, opened 2017-01-30
refactor(Namespaces): implement `.addResource`

silasbw:
`.addResource` adds a resource to the default namespace
(i.e., `api.core.ns.newResource`) and ensures the resource is passed to the
named namespaces (i.e., `api.core.ns('foo').newResource`).  The intended use
case case for `.addResource` is with ThirdPartyResource defined objects.


silasbw:
This is more prep work for https://github.com/godaddy/kubernetes-client/issues/43

jcrugzz:
this LGTM 👍. Clean refactor

PR #70, opened 2017-02-08
feat(ThirdPartyResources): add support for Third Party Resources

### LINKED ISSUE #43: ThirdPartyResource support

silasbw:
Add the ability to create a ThirdPartyResources API object and add resource
objects to it.

Closes #43

### DIFF
--- a/lib/third-party-resources.js
+++ b/lib/third-party-resources.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const ApiGroup = require('./api-group');
+const BaseObject = require('./base');
+
+class ThirdPartyResource extends ApiGroup {
+  constructor(options) {
+    options = Object.assign({}, options, {
+      path: `apis/${ options.group }`,
+      version: options.version || 'v1',
+      genericTypes: []
+    });
+    super(options);
+  }
+
+  addResource(resourceName) {
+    this.namespace.addResource(resourceName);
+    this[resourceName] = new BaseObject({
+      api: this,
+      name: resourceName,
+      parentPath: this.path
+    });
+  }
+}
+
+module.exports = ThirdPartyResource;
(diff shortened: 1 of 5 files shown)


lib/third-party-resources.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const ApiGroup = require('./api-group');
+const BaseObject = require('./base');
+
+class ThirdPartyResource extends ApiGroup {
+  constructor(options) {
```
TWood67: Maybe not for this PR, but it might be a good user experience if one could pass in an array of existing third party resources to the constructor and have the object created here rather than users having to invoke `addResource` on every third party resource themselves.

test/third-party-resources.test.js
```diff
@@ -0,0 +1,117 @@
...
+const assume = require('assume');
+const async = require('async');
+const nock = require('nock');
+
+const common = require('./common');
+const afterTesting = common.afterTesting;
+const beforeTesting = common.beforeTesting;
+const only = common.only;
+
+const thirdPartyDomain = 'kubernetes-client.com';
```
TWood67: I would extract this value to `common.js` since it's used on line 185.

lib/third-party-resources.js
indexzero: Agreed. Something potentially like:

``` js
class ThirdPartyResource extends ApiGroup {
  constructor(options) {
    /* rest of your current code here */
    if (options.resources) {
      options.resources.forEach(res => this.addResource(res));
    }
  }
```

silasbw: Roger that -- I'll add it in a fast follow PR.

PR #71, opened 2017-02-09
feat(ThirdPartyResource): Add `resources` option to constructor

silasbw:


indexzero:
+1

lib/third-party-resources.js
```diff
@@ -20,6 +24,7 @@ class ThirdPartyResource extends ApiGroup {
       name: resourceName,
       parentPath: this.path
     });
+    return this;
```
TWood67: I don't think you need this return

silasbw: Yeah, you're right, but I was trying to sneak in chaining support :)  E.g.,

```js
tpr
  .addResource('foos')
  .addResource('goos')
  .addResource('zoos');
```

PR #73, opened 2017-02-13
feat(config.loadKubeconfig): helper for loading an API config from ~/.kube/config

silasbw:


README.md
```diff
@@ -59,6 +59,13 @@ const K8Api = require('kubernetes-client');
 const k8 = new K8Api.Core(K8Api.config.getInCluster());
 ```
 
+and a helper to get the current-context config from `~/.kube/config`:
+
+```js
+const K8Api = require('kubernetes-client');
+const k8 = new K8Api.Core(K8Api.config.loadKubeconfig());
```
jcrugzz: Should this be fromKubeConfig so it's the right data structure?

silasbw: Oops, yeah, fixed.

lib/config.js
```diff
@@ -2,6 +2,8 @@
 /* eslint no-process-env: 0 no-sync:0 */
 
 const fs = require('fs');
+const path = require('path');
+const yaml = require('js-yaml');
```
jcrugzz: this now needs to be a `dependency` rather than a `devDependency`

silasbw: Good catch, fixed.

Issue #66, opened 2017-01-30
Jobs API endpoint


thataustin:
This may be an inquiry more than a request, but I'm trying to use this library to list/create a job in my cluster.

According to [these docs](https://kubernetes.io/docs/api-reference/batch/v1/operations/), the correct api endpoint is: 
`GET /apis/batch/v1/namespaces/{namespace}/jobs/{name}`.

When I use this library like so (assuming the namespace configured in the k8 constructor is `my-namespace`:
`k8.namespace.jobs('my-job').get(a => console.log(a));`

It makes a request to:

`/api/v1/namespaces/my-namespace/jobs/my-job`

How do I get the first part of that url be `/apis/batch/v1/namespaces`?  Is there a setting I'm missing somewhere?

Thanks, and I appreciate all the work you've done on the library so far!

jcrugzz:
@thataustin this would need https://github.com/godaddy/kubernetes-client/pull/57 to be merged in order to have the correct `batch` `ApiGroup`

thataustin:
Kinda what I figured based on how it seems to be auto-magically constructing the url from the attribute references.

Here's to merging #57 🍺

PR #78, opened 2017-02-18
feat(Support for more APIs): Apps, Batch, and RBAC

silasbw:
Updated version of https://github.com/godaddy/kubernetes-client/pull/57

Closes: https://github.com/godaddy/kubernetes-client/issues/66
Closes: https://github.com/godaddy/kubernetes-client/issues/53

### DIFF
--- a/lib/apps.js
+++ b/lib/apps.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const ApiGroup = require('./api-group');
+
+class Apps extends ApiGroup {
+  constructor(options) {
+    const genericTypes = [
+      // Deprecated name of statefulsets in kubernetes 1.4
+      'petsets',
+      'statefulsets'
+    ];
+    options = Object.assign({}, options, {
+      path: 'apis/apps',
+      version: options.version || 'v1beta1',
+      genericTypes: genericTypes
+    });
+    super(options);
+  }
+}
+
+module.exports = Apps;
--- a/lib/namespaces.js
+++ b/lib/namespaces.js
@@ -12,6 +12,7 @@ const DEFAULT_RESOURCES = [
   { name: 'pods', Constructor: Pods }
 ].concat([
   'configmaps',
+  'cronjobs',
   'daemonsets',
   'deployments',
   'endpoints',
@@ -25,6 +26,8 @@ const DEFAULT_RESOURCES = [
   'petsets',
   'replicasets',
   'resourcequotas',
+  // Deprecated name of cronjobs in kubernetes 1.4
+  'scheduledjobs',
   'secrets',
   'serviceaccounts',
   'services',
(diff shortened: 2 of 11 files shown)


lib/apps.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const ApiGroup = require('./api-group');
+
+class Apps extends ApiGroup {
+  constructor(options) {
+    const genericTypes = [
+      // Deprecated name of statefulsets in kubernetes 1.4
+      'petsets',
```
TWood67: I almost feel it might be worthwhile to have a `deprecatedTypes` array that we can pass to the constructor. Something like:
```
const deprecatedTypes = [ { name: 'petsets', endorsed: 'statefulsets' } ]

options = Object.assign({}, options {
  ...
  deprecatedTypes: deprecatedTypes
});
```

Then in ApiGroup we could have something like:
```
for (const type of options.deprecatedTypes) {
  this[type] = new BaseObject({
    name: type.deprecated,
    parentPath: this.path,
    api: this,
   message: `${ type.deprecated } is a deprecated name. Please use ${ type.endorsed }`
  });
}
```

Then somewhere in the base class we could check for the `message` property and do a `console.warn` with the value in `message` if it exists. Probably in the `CallableObject` class.

Let me know what you think.

jcrugzz: Not a bad idea. The only tricky part is that deprecation depends on what k8s version the user is using. We should be showing a support matrix for types and k8s version after this merges

jcrugzz:
Looks good but needs tests :)

lib/apps.js
TWood67: Good point, my suggestion is premature. I've created #79 so that we don't lose it.

lib/namespaces.js
```diff
@@ -11,7 +11,10 @@ const DEFAULT_RESOURCES = [
   { name: 'replicationcontrollers', Constructor: ReplicationControllers },
   { name: 'pods', Constructor: Pods }
 ].concat([
+  'clusterroles',
```
jcrugzz: Could we keep these strings in one place? we use them for `resourceAliases` as well as keep them in the constructors of the various `ApiGroups`. We should centralize these strings so updating them is not painful.

silasbw: Yeah, but probably out of scope for this PR: https://github.com/godaddy/kubernetes-client/issues/85

silasbw:
I think this is ready to go.  I made:

- https://github.com/godaddy/kubernetes-client/issues/84 for missing RBAC integration tests
- https://github.com/godaddy/kubernetes-client/issues/85 for string copy-pasta

silasbw:
@ankon does this look good to you?

jcrugzz:
LGTM 👍

PR #87, opened 2017-02-25
Use ca.crt correctly

ankon:
The ca.crt file inside a pod is the 'certificate authority', not a 'client certificate'.

---
This should fix errors 'unable to verify the first certificate' when using the in-cluster configuration.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const yaml = require('js-yaml');
 
-const certPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt';
+const caPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt';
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -27,13 +27,13 @@ function getInCluster () {
       ' and KUBERNETES_SERVICE_PORT must be defined');
   }
 
-  const cert = fs.readFileSync(certPath, 'utf8');
+  const ca = fs.readFileSync(certPath, 'utf8');
```
silasbw: rename `certPath` to `caPath`

ankon: Good catch!

I was wondering about the `my-cert` string in the test, but just left it alone -- tell me if that should be changed to `my-ca`, too.

silasbw: yeah, that seems reasonable

silasbw:
Published 3.8.0: https://github.com/godaddy/kubernetes-client/blob/master/CHANGELOG.md#380-2017-02-25

PR #88, opened 2017-03-01
Make check for content-type more generic

njbraun:
Adds support for server responses where the charset is also specified in Content-Type.
For example: 'application/json; charset=utf-8'

### DIFF
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -97,7 +97,7 @@ class ApiGroup {
 
     return request(requestOptions, (err, res, body) => {
       if (err) return cb(err);
-      if (res.headers['content-type'] === 'application/json') {
+      if (res.headers['content-type'].indexOf('application/json') === 0) {
         body = JSON.parse(body);
       }
       cb(null, { statusCode: res.statusCode, body: body });


lib/api-group.js
```diff
@@ -97,7 +97,7 @@ class ApiGroup {
 
     return request(requestOptions, (err, res, body) => {
       if (err) return cb(err);
-      if (res.headers['content-type'] === 'application/json') {
+      if (res.headers['content-type'].indexOf('application/json') === 0) {
```
silasbw: This fails if, for some reason, the server (maybe it's a janky misbehaving HTTP gateway) doesn't add content-type.

The "correct" fix might be to set `json: true` with `request` -- I think this worked poorly with `application/strategic-merge-patch+json` in the past, but I think I remember seeing a change to `request` to accommodate it.  I'd want to add integration tests to ensure it's actually working.

Is this blocking you @njbraun?  If so, fix this PR so it handles the `undefned` content-type and we'll merge it in.  Please add a unit test also.

silasbw:
@njbraun I created https://github.com/godaddy/kubernetes-client/pull/89 to solve this issue.  Thanks for kicking this fix off.

silasbw:
This should be fixed with https://github.com/godaddy/kubernetes-client/pull/89

silasbw:
@njbraun we published the fix as part of version `3.10.0`

PR #89, opened 2017-03-02
fix(response `content-type` checking): more robust JSON response handling

silasbw:
Use `json: true` with request: request has more robust `content-type` checking.


jcrugzz:
👍 if patch merge works correctly

Issue #90, opened 2017-03-08
k8.ns.pvc.matchLabels gives 404


nick:
I used to be able to call the following:

    k8.ns.pvc.matchLabels({ app: 'test' }).get((err, sRes) => {

but that now gives a 'resource not found' error. I now have to call the following instead:

    k8.ns.pvc.get({ qs: { labelSelector: 'app=test' } }, (err, sRes) => {

silasbw:
Ugh, sorry about that.  #91 is a fix.  We do have a test for this (https://github.com/godaddy/kubernetes-client/blob/master/test/api-group.test.js#L208), but it "just happens to still work" because of how we've implemented `Pods` (https://github.com/godaddy/kubernetes-client/blob/master/lib/pods.js#L42).

silasbw:
@nick we published `kubernetes-client@3.10.1` with the fix.  Thanks for the bug report.

PR #91, opened 2017-03-09
fix(matchLabels): call BaseObject constructor with correct arguments

silasbw:
[1] changed BaseObject so that its properties were no longer a
superset of the `options` BaseObject construtor argument (specifically
removing the `name` property). That broke `.match()` and
`.matchLabels()` which depended on that property.

Fixes: #90

[1]:https://github.com/godaddy/kubernetes-client/commit/d083f93be23182f4a629c02cdd319692a09168f8#diff-86b75ecfbd25aa771c8e9e0dff8eb00fL25

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -42,9 +42,11 @@ class BaseObject extends CallableObject {
    *
    * @param {object} options - Options object
    * @param {string} options.api - Kubernetes API URL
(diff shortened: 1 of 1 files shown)


jcrugzz:
LGTM 👍

PR #96, opened 2017-05-11
(fix/feature) Do not override user-provided headers in ApiGroup.patch options, to allow switching patch strategy 

filp:
Kubernetes' PATCH operation accepts different patch strategies:

https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#patch-operations

By default, `kubernetes-client` reasonably uses `application/strategic-merge-patch+json`, however, it's currently impossible to choose a different strategy (which is exactly what I'm trying to do - strategic-merge-patch is problematic in certain situations when dealing with lists):

```js

  /**
   * Invoke a PATCH request against the Kubernetes API server
   * @param {ApiRequestOptions} options - Options object.
   */
  patch(options, cb) {
    this._request('PATCH', Object.assign({}, options, {
      headers: { 'content-type': 'application/strategic-merge-patch+json' }
    }), cb);
  }
```

As you can see, with the current setup, `headers` as merged in the `patch` method ALWAYS takes precedence.

My change reverses the order, allowing the caller to set custom headers, including the patch strategy:

```js
thing.patch({ body: {}, headers: { 'content-type': 'application/merge-patch+json' }, ...)
```

I kept this change to the smallest possible footprint, though you might consider allowing this sort of control at other places of the codebase. It was also unclear where the best place to test this change would be, so I'm open to suggestions.

Cheers!

### DIFF
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -127,7 +127,7 @@ class ApiGroup {
    * @param {callback} cb - The callback that handles the response
    */
   patch(options, cb) {
-    this._request('PATCH', Object.assign({}, {
+    this._request('PATCH', Object.assign({
       headers: { 'content-type': 'application/strategic-merge-patch+json' }
     }, options), cb);
   }
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -127,9 +127,9 @@ class ApiGroup {
    * @param {callback} cb - The callback that handles the response
    */
   patch(options, cb) {
-    this._request('PATCH', Object.assign({}, options, {
+    this._request('PATCH', Object.assign({}, {
       headers: { 'content-type': 'application/strategic-merge-patch+json' }
-    }), cb);
+    }, options), cb);
   }
 
   /**
--- a/lib/base.js
+++ b/lib/base.js
@@ -8,7 +8,7 @@ function cb200(cb) {
     if (result.statusCode < 200 || result.statusCode > 299) {
       const error = new Error(result.body.message || result.body);
       error.code = result.body.code || result.statusCode;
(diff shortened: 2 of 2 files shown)


lib/api-group.js
```diff
@@ -127,9 +127,9 @@ class ApiGroup {
    * @param {callback} cb - The callback that handles the response
    */
   patch(options, cb) {
-    this._request('PATCH', Object.assign({}, options, {
+    this._request('PATCH', Object.assign({}, {
```
silasbw: You can drop the empty object to `Object.assign`.  _I.e._,:

```js
this._request('PATCH', Object.assign({
  headers: { 'content-type': 'application/strategic-merge-patch+json' }
}, options), cb);
```

lib/base.js
```diff
@@ -8,7 +8,7 @@ function cb200(cb) {
     if (result.statusCode < 200 || result.statusCode > 299) {
       const error = new Error(result.body.message || result.body);
       error.code = result.body.code || result.statusCode;
-      return cb(error);
+      return cbba(error);
```
filp: Yep, seems like I fat-fingered it - will fix ASAP.

lib/api-group.js
filp: Good catch, will do that as well.

filp:
@silasbw Thanks for having a look and the initial feedback. I'll have a look at implementing this across other methods. Cheers!

silasbw:
Hey @filp, I wanted to get these improvements in, so i went ahead and added the missing tests -- see https://github.com/godaddy/kubernetes-client/pull/107

Issue #47, opened 2016-12-16
Support more Deployment operations


silasbw:
E.g., POST /apis/extensions/v1beta1/namespaces/{namespace}/deployments/{name}/rollback

This is a special case of a more general problem we need to handle with other objects (e.g., `proxy` for Pods and Services).

PR #97, opened 2017-05-13
refactor(ApiGroup): support custom root group resources

silasbw:
The current code assumes every root group resources (e.g., /api/v1/nodes) is a
BaseObject. This adds support for custom resources (e.g., Pod or Deployment).
This is groundwork for:

https://github.com/godaddy/kubernetes-client/issues/47

### DIFF
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -62,7 +62,7 @@ class ApiGroup {
   addResource(options) {
     if (typeof options === 'string') {
       options = { name: options, Constructor: BaseObject };
-    } else if (!option.name || !options.Constructor) {
+    } else if (!options.name || !options.Constructor) {
       throw new RangeError(
         'ApiGroup.addResource: options requires .name and .Constructor');
     }
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -62,6 +62,14 @@ class ApiGroup {
   addResource(options) {
     if (typeof options === 'string') {
       options = { name: options, Constructor: BaseObject };
+    } else if (!option.name || !options.Constructor) {
+      throw new RangeError(
+        'ApiGroup.addResource: options requires .name and .Constructor');
+    }
+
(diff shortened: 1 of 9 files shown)


lib/api-group.js
```diff
@@ -46,19 +46,31 @@ class ApiGroup {
...
     aliasResources(this);
   }
 
   /**
+   * Add a resource (e.g., nodes) to the ApiGroup group
+   * @param {string|object} options - resource name or options object
+   * @param {string} options.name - resource name
+   * @param {fn} options.Constructor - constructor for new resource
+   */
+  addResource(options) {
```
satish-ravi: Should we validate options to ensure it includes `name` and `Constructor` if it is an object?

lib/api-group.js
```diff
@@ -46,19 +46,31 @@ class ApiGroup {
...
+    if (typeof options === 'string') {
+      options = { name: options, Constructor: BaseObject };
+    }
+    this[options.name] = new options.Constructor({
```
satish-ravi: Is it ok to override resource if one already exists?

silasbw: It's probably something to avoid doing.  I'll add a check to enforce that.

lib/api-group.js
```diff
@@ -46,19 +46,39 @@ class ApiGroup {
...
   /**
+   * Add a resource (e.g., nodes) to the ApiGroup group
+   * @param {string|object} options - resource name or options object
+   * @param {string} options.name - resource name
+   * @param {fn} options.Constructor - constructor for new resource
+   */
+  addResource(options) {
+    if (typeof options === 'string') {
+      options = { name: options, Constructor: BaseObject };
+    } else if (!option.name || !options.Constructor) {
```
satish-ravi: `options.name`. Should we add a unit test for missing name/Constructor?

PR #99, opened 2017-05-14
refactor(linter): Upgrade godaddy-style

silasbw:
* Do a lint pass; and
* add lint checks to the test script.


silasbw:
See: https://github.com/godaddy/kubernetes-client/pull/100

PR #102, opened 2017-05-14
refactor(BaseObject): add a `ContainerBaseObject` to facilitate named resources

silasbw:
Some named resources (e.g., `pod('foo'))` can contain resources (e.g.,
`pod('foo').log` or `deploy('bar').status`). The ContainerBaseObject makes
implementing these resources easier.

This is groundwork for https://github.com/godaddy/kubernetes-client/issues/47.

### DIFF
--- a/lib/container-base-object.js
+++ b/lib/container-base-object.js
@@ -1,0 +1,49 @@
+'use strict';
+
+const BaseObject = require('./base');
+
+class ContainerBaseObject extends BaseObject {
+  /**
+   * Create generic Kubernetes API object that might contain other resources.
+   * For example, a named Pod contains .log resources (core.ns.pods('foo').log).
+   *
+   * @param {object} options - Options object
+   * @param {string} options.resources - Array of resources to add
+   */
+  constructor(options) {
+    super(options);
+    if (options.resources) {
+      options.resources.forEach(resource => this.addResource(resource));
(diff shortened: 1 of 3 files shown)


lib/container-base-object.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   * Add a resource to the container object.
+   * @param {string|object} options - resource name or options object
+   * @param {string} options.name - resource name
+   * @param {fn} options.Constructor - constructor for new resource
+   * @returns {object} returns this to facilitate chaining
+   */
+  addResource(options) {
```
satish-ravi: This looks the same as the one in `ApiGroup`. Any way we can reuse?

silasbw: I tried several different ways to reuse, but wasn't super pleased with the results.  Do you have a suggestion?  They're different enough (e.g., `this` for ApiGroup vs `this.api` for `ContainerBaseObject`) that it's pretty annoying.  I think eventually `ApiGroup` should become a `ContainerBaseObject` and then we can nuke the duplicate in `ApiGroup`.

satish-ravi: Not really. I think it is ok to duplicate if we are going to eventually make `ApiGroup` a `ContainerBaseObject`

PR #107, opened 2017-06-01
feat(Support for user-specified PATCH operations)

silasbw:
Allows specifying these patch opertions: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#patch-operations

### DIFF
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -148,9 +148,9 @@ class ApiGroup {
    * @param {callback} cb - The callback that handles the response
    */
   patch(options, cb) {
-    this._request('PATCH', Object.assign({}, options, {
+    this._request('PATCH', Object.assign({}, {
       headers: { 'content-type': 'application/strategic-merge-patch+json' }
-    }), cb);
+    }, options), cb);
   }
 
   /**
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -159,9 +159,9 @@ class ApiGroup {
    * @param {callback} cb - The callback that handles the response
    */
   post(options, cb) {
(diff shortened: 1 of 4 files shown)


lib/api-group.js
```diff
@@ -148,9 +148,9 @@ class ApiGroup {
    * @param {callback} cb - The callback that handles the response
    */
   patch(options, cb) {
-    this._request('PATCH', Object.assign({}, options, {
+    this._request('PATCH', Object.assign({
       headers: { 'content-type': 'application/strategic-merge-patch+json' }
-    }), cb);
+    }, options), cb);
```
jcrugzz: should we do a recursive merge so that headers that arent `content-type` can be specified? Not sure if thats useful but thought I'd ask.

TWood67: I agree. The non-recursive strategy could cause someone grief

lib/base.js
```diff
@@ -120,8 +120,12 @@ class BaseObject extends CallableObject {
    * @param {callback} cb - The callback that handle the response
    */
   patch(options, cb) {
-    this.api.patch({ path: this._path(options), body: options.body },
-                   cb200(cb));
+    const patchOptions = {
+      path: this._path(options),
+      body: options.body
+    };
+    if ('headers' in options) patchOptions.headers = options.headers;
```
TWood67: I don't think this check is needed... The request options here will grab headers regardless if they exist or not
https://github.com/godaddy/kubernetes-client/blob/a4708cc24f6d29930d32aa6d12b4e07885c42765/lib/api-group.js#L109-L116
Might as well just assign the value without checking for the property.

PR #122, opened 2017-06-30
feat(Deployments): Support apps/deployments

gempesaw:
Hey, are you interested in a PR for adding deployments under `apps` ? If so, I'll see about adding tests and all to this as well instead of just a one-liner. It looks like [apps/v1beta1.Deployment](https://github.com/kubernetes/kubernetes.github.io/pull/3814#issuecomment-303239179) "will eventually be the default," and is working that way in 1.6, but for the time being deployments are under both extensions and apps. 

I've only just started using this client library, and I know it's hard to support multiple versions at once, so I understand if you've got other concerns that would discourage this PR. Thanks!

### DIFF
--- a/lib/apps.js
+++ b/lib/apps.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const ApiGroup = require('./api-group');
+const Deployments = require('./deployments');
 
 class Apps extends ApiGroup {
   constructor(options) {
@@ -12,8 +13,10 @@ class Apps extends ApiGroup {
     options = Object.assign({}, options, {
       path: 'apis/apps',
       version: options.version || 'v1beta1',
-      groupResources: resources,
-      namespaceResources: resources
+      groupResources: resources.concat(['deployments']),
+      namespaceResources: resources.concat([
+        { name: 'deployments', Constructor: Deployments }
+      ])
     });
     super(options);
   }
--- a/test/apps.test.js
+++ b/test/apps.test.js
@@ -34,6 +34,32 @@ const testStatefulSet = {
   }
 };
 
+const testDeployment = {
+  apiVersion: 'apps/v1beta1',
+  kind: 'Deployment',
+  metadata: {
+    name: 'web'
+  },
+  spec: {
+    template: {
+      spec: {
+        containers: [
+          {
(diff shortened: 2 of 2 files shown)


silasbw:
Hey @gempesaw, sorry for the delay -- a bunch of us were on vacation.

Thank you for submitting this and yes, we should support apps/Deployments.  See https://github.com/godaddy/kubernetes-client/blob/master/lib/extensions.js#L24 for how to support Deployments.

gempesaw:
cool! i'm planning to rebase to pick up the changes in master (just saw a new button in the github UI to pick up new changes and I wanted to see what it would do!), and then i'll fix the change and see if there are any tests for analogous stuff that i can copy & tweak for apps/Deployments. 

no worries about the delays, vacation is great! and i'm way slower on my own projects, lol...

gempesaw:
I finally found time for this again, a month after I said I'd do it! :D

I rebased on top of the latest master, adding deployments to the namespaceResources as you suggested and duplicating the `lib.apps.statefulsets` tests for deployments. As expected, this enables things like 

```node
// all async, this won't run in order...
api.apps.ns.deployments.post({ body: deployManifest }, print); 
api.apps.ns.deployments.put({ body: deployManifest, name: deployManifest.metadata.name }, print);
api.apps.ns.deployments(deployManifest.metadata.name).get(print);
api.apps.ns.deployments.delete(deployManifest.metadata.name, print);
```

and the helper functionality as well:

```node
api.group(deployManifest).ns.kind(deployManifest).post({ body: deployManifest }, print);
api.group(deployManifest).ns.kind(deployManifest).delete({ name: deployManifest.metadata.name }, () => {});
```

I noticed that your example file also had `['deployments']` concatted to the groupResources, I couldn't figure out what that would do for me, but since the tests and the above function chains worked fine, I didn't include it. 

Feel free to let me know if I should change anything, and thanks again!

gempesaw:
Ahh, I haven't used that endpoint before, but that makes sense. Added and force-pushed again, thanks :)

silasbw:
...published this change to npm as 3.14.0, thanks again @gempesaw.

Issue #37, opened 2016-10-28
Promises based API


kjvalencik:
It would be nice if this library also supported promises. Unfortunately, wrapping the `request` library for promises is fairly [non-trivial](https://github.com/request/promise-core/blob/master/lib/plumbing.js). This makes it a bit unwieldy to promisify at runtime.

The [request-promise](https://github.com/request/request-promise) library also supports a `nodeify` method for callbacks, allowing simulataneous support for callbacks and promises. However, this may make streaming more cumbersome.

Perhaps a provide your own `request` library approach?

silasbw:
Sorry for the delay @kjvalencik -- this is something we'd consider implementing, but it hasn't been high on our priority list because we usually eschew promises.

I think the first step would be rejiggering kubernetes-client a little bit to facilitate adding promise support.  If you have some concrete thoughts here, we'd love to hear them.

Second step would then be actually adding the promise support.  Either as an option to kubernetes-client, or as a separate `kubernetes-client-promises` wrapper package.

kjvalencik:
There are a few things that I think could be helpful to allow integrating promises externally:

* Ensure all calls return their result. This is also useful if you want to stream the response.
* Optional config to pass the `request` library in.
* There are a few places in the code that don't check that a parameter exists before grabbing a value off it. These work because a `callback` is usually passed. Without it you will get an `undefined` deref error.
* Pass `json: true` to `request` so that parsing happens in the library instead of in a callback.

Open questions:

* How to implement retry logic that works regardless of `request` library?
* What to do with methods that do work in a wrapping callback before handing off to a user callback?
* How to test to prevent regressions in promise functionality?

silasbw:
@kjvalencik helpful list -- if we could rejigger things so that switching to promise mode was as easy as passing in the `request` library, that would be great.  I'd worry about making the `kubernetes-client` code opaque and cumbersome to maintain, but that's a qualitative thing that would be hard to evaluate without looking at a PR first.

re: `json: true` -- there's reason we didn't do that to begin with and it's that when set and `content-type: application/strategic-merge-patch+json`, `request` (essentially) ignores `json: true` so our client code ends up implementing the JSON decoding anyways.  To be consistent we just don't set it.  We've meant to address this issue via a PR to `request`, but haven't done it yet.

silasbw:
We should consider adding a promise-based API to kubernetes-client:

```js
const api = require('kubernetes-client/async');
[...]

PR #126, opened 2017-07-09
feat(Promise API): Experimental support for a promise-based API

silasbw:
This preserves the original callback-based "core" implementation and layers the
promised-based API on top of that core implementation. The goal is start using
and iterating on the promised-based API even if we expect the underlying
implementation to evolve (e.g., we could consider replacing the core
callback-based implementation with a promise one).  See
https://github.com/godaddy/kubernetes-client/issues/37 for discussion.

Example usage:

```js
async function main() {
  try {
    const pods = await core.ns('kube-system').po.get();
    console.log(`Watching: ${ JSON.stringify(pods, null, 2) }`);
    const stream = core.ns('kube-system').po.getStream({ qs: { watch: true }});
    stream.on('data', data => { console.log(data.toString()); });
    stream.on('error', err => { throw err; });
    stream.on('end', () => { console.log('end'); });
  } catch (error) {
    console.log(error);
  }
}
main();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,39 @@ const Api = require('kubernetes-client');
 const core = new Api.Core(Api.config.fromKubeconfig());
 ```
 
+### **Experimental** support for promises and async/await
+
+kubernetes-client exposes **experimental** support for promises via
+the `kubernetes-client/promise` path. The API is the same, except for the
+functions that previously took a callback (*e.g.*, `.get`). Those functions now
+return a promise.
+
+```js
+// Notice the require path...
+const Api = require('kubernetes-client/promise');
+// ...but the same constructor arguments.
+const core = new Api.Core({
+  url: 'http://my-k8s-api-server.com',
+  version: 'v1',  // Defaults to 'v1'
+  namespace: 'my-project' // Defaults to 'default'
+});
+```
+
+and then:
+
+```js
+core.namespaces.replicationcontrollers('http-rc').get()
+  .then(result => JSON.stringify(result, null, 2))
+  .catch(err => throw err);
+```
+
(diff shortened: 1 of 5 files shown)


indexzero:
Why not encourage folks to use the util.promisify API in Node.js itself?

silasbw:
Good question.  Three pretty mushy reasons:

1. Some interest from users for promise-based API (#37)
2. Personal preference -- I'd rather do `await core.ns.po.get()` than `await promisify(core.ns.po.get)()`
3. If we expect async/await usage to increase, creating a promise-based API now could be a good long term investment towards supporting the async/await users (if they share my personal preferences)

It seems like there's some pros to adding this, and it's a pretty low cost con for maintaining this code.  Are there bigger cons you're worried about?

indexzero:
Mainly browser (i.e. `webpack` compatibility) of these changes, but I'm not sure that's in the scope of the requirements you're trying to reach. Another way to look at this is that if the goal is to make `async / await` more easily accessible: is that just server-side? Or do we want to care about polyfill targets, such as `Proxy`?

silasbw:
Oh, that makes sense.  We're currently incompatible with browsers (*e.g.*, dependencies on `request`), but I'd like to be.  I also realize that if we want to support Node.js 4.X we run into the same problem with `Proxy` (I'd also need to clean up the spread operator uses).

I was hoping for an "easy" promise implementation, but it seems like `Proxy` usage might cause some issues.  I might be able to eliminate `Proxy` without adding too much complexity.

silasbw:
Closing in favor of https://github.com/godaddy/kubernetes-client/pull/133

PR #133, opened 2017-07-22
feat(Promise API): Experimental support for a promise-based API

silasbw:
This preserves the original callback-based "core" implementation and layers the
promised-based API on top of that core implementation. The goal is start using
and iterating on the promised-based API even if we expect the underlying
implementation to evolve (e.g., we could consider replacing the core
callback-based implementation with a promise one).  See
https://github.com/godaddy/kubernetes-client/issues/37 for discussion.

Example usage:

```js
const core = new api.Core({
  url: 'http://my-k8s-api-server.com',
  promises: true
});

async function main() {
  try {
    const pods = await core.ns('kube-system').po.get();
    console.log(`Watching: ${ JSON.stringify(pods, null, 2) }`);
    const stream = core.ns('kube-system').po.getStream({ qs: { watch: true }});
    stream.on('data', data => { console.log(data.toString()); });
    stream.on('error', err => { throw err; });
    stream.on('end', () => { console.log('end'); });
  } catch (error) {
    console.log(error);
  }
}
main();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,43 @@ const Api = require('kubernetes-client');
 const core = new Api.Core(Api.config.fromKubeconfig());
 ```
 
+### **Experimental** support for promises and async/await
+
+kubernetes-client exposes **experimental** support for promises via
+the `promises` option passed to API group constructors. The API is the
+same, except for the functions that previously took a callback
+(*e.g.*, `.get`). Those functions now return a promise.
+
+```js
+// Notice the promises: true
+const core = new Api.Core({
+  url: 'http://my-k8s-api-server.com',
+  version: 'v1',  // Defaults to 'v1'
+  promises: true,  // Enable promises
+  namespace: 'my-project' // Defaults to 'default'
+});
+```
+
+and then:
+
+```js
+core.namespaces.replicationcontrollers('http-rc').get()
+  .then(result => print(null, result));
+```
+
+or with `async/await`:
+
+```js
+print(null, await core.namespaces.replicationcontrollers('http-rc').get());
+```
+
+You can invoke promise-based and callback-based functions explictly:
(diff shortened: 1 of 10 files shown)


kilianc:
I was about to open an issue and start a PR on this. Thanks!!!!

If I may, why `promises: true` versus supporting callback *and* returning a promise?

jackzampolin:
Has work on this stalled? Is this feature working on the branch?

silasbw:
I can give an update.

1. There was some concern around the performance impact of these `bind`s: https://github.com/godaddy/kubernetes-client/pull/133/files#diff-86b75ecfbd25aa771c8e9e0dff8eb00fR79
2. I've been waffling on the API (*i.e.*, @kilianc's comment https://github.com/godaddy/kubernetes-client/pull/133#issuecomment-326352272 is a good one).  kubernetes-client already overloads `.get` to return a stream if no callback is passed (this isn't documented and we discourage it, but it used to be documented and encouraged).  If we implemented @kilianc's comment we'd break the users that still do this, and if we bumped major versions we make it more cumbersome for them to upgrade.

Regarding:

1. We have a fairly large implementation change coming that will make this point moot (check out @jeffpearce's work to automatically generate API bindings https://github.com/godaddy/kubernetes-client/tree/jxpearce/swag/api-gen)
2. My waffling has been mostly unproductive, but I think we should roll this out with as is.  Longer term, if overloading `.get`, `.post`, etc is something we think would be helpful and the old stream API users are in the minority, then we should implement the new overloaded behavior.

(and yes, this branch is working...at least according to the unit and integration tests :smile: )

silasbw:
FYI, I just published 3.16.0 and included this feature.

PR #142, opened 2017-08-22
Make possible to send body in delete requests

anton-kotenko:
Some kubernetes api's requires to send body within delete request.
Examples: [job deletion](https://kubernetes.io/docs/api-reference/v1.7/#delete-41) and [pod deletion](https://kubernetes.io/docs/api-reference/v1.7/#delete-58)
I needed to send DeleteOptions in request body to make job deletion also to delete all related pods.

Changes in this pull request provides possibility to do this in same way, as it works for POSTs.
````
const K8s = require('kubernetes-client');
const batch = new K8s.Batch({/*add configuration*/});
batch.namespaces('default').jobs.delete({
    name: 'some-job',
    body: {
        kind: 'DeleteOptions',
        apiVersion: 'batch/v1',
        propagationPolicy: 'Foreground'
    }
}, (err, data) => console.log(err, data))

````

### DIFF
--- a/test/base.test.js
+++ b/test/base.test.js
@@ -6,6 +6,7 @@ const async = require('async');
 const nock = require('nock');
 
 const common = require('./common');
+const only = common.only;
 const beforeTesting = common.beforeTesting;
 
 function pod(name) {
@@ -88,6 +89,30 @@ describe('lib.base', () => {
           assume(err).is.falsy();
           assume(pods.kind).is.equal('PodList');
           assume(pods.items).has.length(1);
+          done();
+        });
+      });
+    });
+
+    describe('.delete', () => {
+      const podName = 'pod-name';
+      const query = { pretty: true };
+      const body = {
+        apiVersion: 'v1',
+        kind: 'Pod',
+        propagationPolicy: 'Foreground'
+      };
+      beforeTesting('unit', () => {
+        nock(common.api.url)
+          .delete(`/api/v1/namespaces/${ common.currentName }/pods/${ podName }`, body)
+          .query(query)
+          .reply(200, { kind: 'Pod' });
+      });
+
+      only('unit', 'should bypass query string and body from arguments into request', done => {
+        common.api.ns.po.delete({ name: podName, qs: query, body: body }, (err, result) => {
+          assume(err).is.falsy();
+          assume(result).is.eql({ kind: 'Pod' });
           done();
         });
       });
--- a/test/base.test.js
+++ b/test/base.test.js
@@ -6,7 +6,6 @@ const async = require('async');
 const nock = require('nock');
 
 const common = require('./common');
-const only = common.only;
 const beforeTesting = common.beforeTesting;
 
 function pod(name) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by drobiazko** [comment]
Guys, any feedback on this pull request? Do you think it can be merged soon?

File: test/base.test.js [hunk 1]
```diff
@@ -93,6 +94,30 @@ describe('lib.base', () => {
...
+        propagationPolicy: 'Foreground'
+      };
+      beforeTesting('unit', () => {
+        nock(common.api.url)
+          .delete(`/api/v1/namespaces/${ common.currentName }/pods/${ podName }`, body)
+          .query(query)
+          .reply(200, { kind: 'Pod' });
+      });
+
+      only('unit', 'should bypass query string and body from arguments into request', done => {
```
TWood67: Could you add support for integration tests as well? I'd add a little more validation to your tests as well, like verifying that the body passed in performed the expected operation on the cluster.

test/base.test.js
```diff
@@ -93,6 +94,30 @@ describe('lib.base', () => {
       });
     });
 
+    describe('.delete', () => {
+      const podName = 'pod-name';
+      const query = { pretty: true };
+      const body = {
+        apiVersion: 'v1',
+        kind: 'Pod',
```
silasbw: I think this should be `'DeleteOptions'` (it's the Kind of object you're passing in the body, not the Kind of object you want to delete)

test/base.test.js
```diff
@@ -93,6 +94,30 @@ describe('lib.base', () => {
...
+        propagationPolicy: 'Foreground'
+      };
+      beforeTesting('unit', () => {
+        nock(common.api.url)
+          .delete(`/api/v1/namespaces/${ common.currentName }/pods/${ podName }`, body)
+          .query(query)
+          .reply(200, { kind: 'Pod' });
+      });
+
```
silasbw: Add something like:

```js
beforeTesting('int', done => {
    common.api.ns.po.post({ body: pod(podName) }, done);
});
```

test/base.test.js
silasbw: If you make the addition I suggest above, change this to a simple `it` to support integration tests:

```js
it('should bypass query string and body from arguments into request', done => {
```

test/base.test.js
```diff
@@ -93,6 +94,30 @@ describe('lib.base', () => {
...
+      only('unit', 'should bypass query string and body from arguments into request', done => {
+        common.api.ns.po.delete({ name: podName, qs: query, body: body }, (err, result) => {
+          assume(err).is.falsy();
+          assume(result).is.eql({ kind: 'Pod' });
```
silasbw: You'll need to make this more general to handle integration tests.  I suggest something simple just to ensure we're actually piping the delete response through correctly.  E.g.,

```js
assume(result.kind).is.equal('Pod');
```

silasbw:
Also, sorry for the delay -- it's vacation season for many of us.

anton-kotenko:
@TWood67, @silasbw 
Thank you for feedback on pull request.
Did the things, you've suggested, that is mostly adding integration tests.

gempesaw:
just wanted to say thanks for the PR and merging in the changes! serendipitously, I was looking for this exact functionality on the same day it was merged :D

PR #146, opened 2017-09-17
Allow to overwrite `path`

Schnitzel:
First: Thanks for this :)

I'm using kubernetes-clienta against an openshift and some of the apis of OpenShift are at `/oapi/` instead of `/api/`

This patch allows to do:
```
const openshift = new Api.Core({
  url: 'https://192.168.99.105:8443',
  path: 'oapi',
});
```

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -20,7 +20,7 @@ class Core extends ApiGroup {
       'services'
     ];
     options = Object.assign({}, options, {
-      path: 'api',
+      path: options.path || 'api',
       version: options.version || 'v1',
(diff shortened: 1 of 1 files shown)


silasbw:
hey @Schnitzel, thanks for the PR.  Can you help me understand how you might use this?  Won't you need access to additional properties on the `openshift` object (in your example above) for this be be useful?  E.g., https://github.com/debianmaster/openshift-client/blob/master/lib/oapi.js

silasbw:
hey @Schnitzel is this still relevant?

PR #152, opened 2017-10-03
Added an autoscaling api to enable HorizontalPodAutoscaler Api calls

dackers86:


lib/autoscaling.js
```diff
@@ -0,0 +1,23 @@
...
+
+class Autoscaling extends ApiGroup {
+  constructor(options) {
+    const commonResources = [
+      'horizontalpodautoscalers'
+    ];
+    options = Object.assign({}, options, {
+      path: 'apis/autoscaling',
+      version: options.version || 'v1',
+      groupResources: commonResources.concat(['deployments']),
```
silasbw: You'd only need `deployments` here if there was a `/apis/autoscaling/v1/deployments` resource, which there isn't (nor is there a `/apis/autoscaling/v1/namespace/:namespace/deployments` resource.

Remove the deployments.

test/autoscaling.test.js
```diff
@@ -0,0 +1,135 @@
...
+                .get(resourcePath)
+                .reply(200, mockDs)
+                .delete(resourcePath)
+                .reply(200, mockDs);
+    });
+
+    it('POSTs, GETs, and DELETEs', done => {
+      async.series([
+        next => {
+          console.log('here', common.autoscaling.ns.ds);
```
silasbw: remove

test/autoscaling.test.js
```diff
@@ -0,0 +1,135 @@
...
+          }
+        };
+
+        nock(common.autoscaling.url)
+                    .get(path)
+                    .reply(200, mockDsList);
+      });
+
+      it('returns DaemonSetList', done => {
+        async.series([
```
silasbw: the `async.series` is superfluous here, right?  Refactor to remove it if so.

test/autoscaling.test.js
```diff
@@ -0,0 +1,135 @@
...
+
+const common = require('./common');
+const beforeTesting = common.beforeTesting;
+const resourceName = 'test-name';
+
+describe('lib.autoscaling', () => {
+  describe('.autoscaling', () => {
+    const path = `/apis/autoscaling/v1/namespaces/${ common.currentName }/horizontalpodautoscalers`;
+    const resourcePath = `${ path }/${ resourceName }`;
+    const horizontalAutoscaleObj = {
```
silasbw: An almost identical value for a `HorizontalPodAutoscaler` resource is defined several times (*e.g*., https://github.com/godaddy/kubernetes-client/pull/152/files#diff-179fe5b2cb9d0365ec10002d994f44d2R41 and https://github.com/godaddy/kubernetes-client/pull/152/files#diff-179fe5b2cb9d0365ec10002d994f44d2R92).

Can you reduce that duplication?

test/autoscaling.test.js
```diff
@@ -0,0 +1,135 @@
...
+            subresource: 'scale'
+          }
+        },
+        status: {
+          currentReplicas: 0,
+          desiredReplicas: 0
+        }
+      };
+      nock(common.autoscaling.url)
+                .post(path)
```
silasbw: This indent looks off.  I would have expected two spaces from `nock` statement:

```js
nock(common.autoscaling.url)
  .post(path)
  .reply(201, mockDs)
  ...
```

Can you adjust it?

I'm also surprised `npm run lint` or `npm run test` didn't complain, but that's a different matter.

test/autoscaling.test.js
dackers86: Duplicates removed...

test/autoscaling.test.js
dackers86: Formatting updated, npm lint shows the following output...

```
/usr/local/bin/node /usr/local/lib/node_modules/npm/bin/npm-cli.js run lint --scripts-prepend-node-path=auto

> kubernetes-client@3.16.0 lint /Users/darrenackers/Projects/ThirdParty/kubernetes-client
> eslint-godaddy lib/ test/


Process finished with exit code 0
```

Potentially an issue with my machine?

test/autoscaling.test.js
dackers86: Done

test/autoscaling.test.js
dackers86: Removed in extensions and autoscaling tests.

lib/autoscaling.js
silasbw: Yikes, that's unexpected.  I think this works for me:

```js
'use strict';

const ApiGroup = require('./api-group');

class Autoscaling extends ApiGroup {
  constructor(options) {
    const commonResources = [
      'horizontalpodautoscalers'
    ];
    options = Object.assign({}, options, {
      path: 'apis/autoscaling',
      version: options.version || 'v1',
      groupResources: commonResources,
      namespaceResources: commonResources
    });
    super(options);
  }
}

module.exports = Autoscaling;
```

test/autoscaling.test.js
```diff
@@ -0,0 +1,84 @@
...
+describe('lib.autoscaling', () => {
+  describe('.autoscaling', () => {
+    const path = `/apis/autoscaling/v1/namespaces/${ common.currentName }/horizontalpodautoscalers`;
+    const resourcePath = `${ path }/${ resourceName }`;
+    const horizontalAutoscaleObj = {
+      apiVersion: 'autoscaling/v1',
+      kind: 'HorizontalPodAutoscaler',
+      metadata: {
+        name: 'test-name',
+        namespace: 'test-namespace'
```
silasbw: Hard-coding the namespace here will cause integration tests to fail.  I think you can remove this line without any ill effects and also fix integration tests.

test/autoscaling.test.js
```diff
@@ -0,0 +1,84 @@
...
+      beforeTesting('unit', () => {
+        nock(common.autoscaling.url)
+        .get(path)
+        .reply(200, horizontalAutoscaleObj);
+      });
+
+      it('returns HorizontalPodAutoscalerList', done => {
+        common.autoscaling.ns.hpa.get((err, result) => {
+          assume(err).is.falsy();
+          assume(result.kind).is.equal('HorizontalPodAutoscaler');
```
silasbw: This should be a `HorizontalPodAutoscalerList`, yeah?

dackers86: Yes, added an update the test data to be hpa (Horizontal Pod Autoscaler), rather than ds (daemonsets)

test/autoscaling.test.js
dackers86: Namespace removed

lib/autoscaling.js
dackers86: Updated with code sample...

test/autoscaling.test.js
silasbw: Sorry, I meant that `common.autoscaling.ns.hpa.get((err, result) => {` is going to call the callback with a result of `kind` `HorizontalPodAutoscalerList`, not `HorizontalPodAutoscaler`.  I caught this by running the integration tests locally.  I think you need to update the `nock`ing to return a `HorizontalPodAutoscalerList`; and then change the assume statement to be `assume(result.kind).is.equal('HorizontalPodAutoscalerList');`

silasbw:
Hey, thanks for this.  It's no supported / covered by the kubernetes-client generated by openapi.

PR #153, opened 2017-10-24
fix: accept relative paths

raszi:
Our configuration uses relative certificate settings. It is working very well with the `kubectl` command but `kubernetes-client` does not support relative certificates.

Unfortunately the current API does not work very well with this setup because the path of the current config file is not available only in the default case. I fixed the default config handling to make it work but I know it isn't entirely right.

I am open to other solutions or recommendations.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -9,6 +9,11 @@ const caPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt';
 const tokenPath = '/var/run/secrets/kubernetes.io/serviceaccount/token';
 const namespacePath = '/var/run/secrets/kubernetes.io/serviceaccount/namespace';
 
+function defaultConfigPath() {
+  const homeDir = process.env[(process.platform === 'win32') ? 'USERPROFILE' : 'HOME'];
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -9,6 +9,11 @@ const caPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt';
 const tokenPath = '/var/run/secrets/kubernetes.io/serviceaccount/token';
 const namespacePath = '/var/run/secrets/kubernetes.io/serviceaccount/namespace';
 
+function defaultConfigPath() {
```
raszi: I couldn't decide whether you prefer to have `homeDir` and `defaultConfig` as constants or as a function as it is now.

raszi:
I came up with a better idea, I'll move the file reading part to the `loadKubeconfig` function and then it would work independently of the config file location.

If you would like complete backward compatibility then I could also keep the file reading in the `fromKubeconfig` if somebody constructs the config by themself.

raszi:
I created two backward compatible version with #154 and #155 which is a better solution for the problem.

raszi:
Closing in favor of #155

PR #154, opened 2017-10-31
fix: accept relative paths

raszi:
Mutable version, Closes #153, #155


raszi:
Closing in favor of #155

PR #155, opened 2017-10-31
fix: accept relative and absolute paths

raszi:
Immutable version, Closes #153, #154

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -120,7 +120,10 @@ module.exports.fromKubeconfig = fromKubeconfig;
 
 function normalizePathFn(cfgPath) {
   const configDir = path.dirname(cfgPath);
-  return (subPath) => path.isAbsolute(subPath) ? subPath : path.join(configDir, path.normalize(subPath));
+  return (subPath) => {
+    const normalizedPath = path.normalize(subPath);
+    return path.isAbsolute(normalizedPath) ? normalizedPath : path.join(configDir, normalizedPath);
+  };
 }
 
 function mapSpecificKeysFn(mapFn, srcKey) {
--- a/lib/config.js
+++ b/lib/config.js
@@ -9,6 +9,11 @@ const caPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt';
 const tokenPath = '/var/run/secrets/kubernetes.io/serviceaccount/token';
 const namespacePath = '/var/run/secrets/kubernetes.io/serviceaccount/namespace';
 
+function defaultConfigPath() {
+  const homeDir = process.env[(process.platform === 'win32') ? 'USERPROFILE' : 'HOME'];
+  return path.join(homeDir, '.kube', 'config');
+}
+
 /**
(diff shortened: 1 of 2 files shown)


silasbw:
Thanks for this @raszi!  kubernetes-client definitely needs to handle relative paths in the kubeconfig.  What do you think about handling that here:

https://github.com/godaddy/kubernetes-client/blob/master/lib/config.js#L80

At that line kubernetes-client would detect if it's a relative path or an absolute one and operate accordingly.

Thoughts?

raszi:
Well, the problem with that solution is that we don't have the path of the config file there only the config structure.

So what I could do - and this makes sense - is to not read the certificates in the `loadKubeconfig` function but change the path to absolute and then the original code could read the certs.

raszi:
I can make these changes but could you please decide which solution do you prefer @silasbw the mutable or the immutable one?

silasbw:
> Well, the problem with that solution is that we don't have the path of the config file there only the config structure.

Doh, you're right.

> So what I could do - and this makes sense - is to not read the certificates in the loadKubeconfig function but change the path to absolute and then the original code could read the certs.

This sounds good :medal_sports: !

raszi:
Allright, could you decide which PR should I refactor #154 or #155?

silasbw:
#155 seems reasonable.

raszi:
@silasbw I believe this is done.

lib/config.js
```diff
@@ -113,12 +118,45 @@ function fromKubeconfig(kubeconfig, current) {
...
+function mapSpecificKeysFn(mapFn, srcKey) {
+  return (source) => {
+    return Object.keys(source).reduce((dest, key) => {
+      dest[key] = (key === srcKey) ? mapFn(source[key]) : source[key];
+      return dest;
+    }, {});
+  };
+}
+
+function mapCertificates(cfgPath, config) {
```
silasbw: I appreciate the work to implement the helpers / map functions, but it took me a while to understand it completely.  I'd worry it might give others who wish to extend or maintain it pause also.  What about a simple-stupid implementation?  E.g.,

```js
config.clusters.filter(cluster => cluster.cluster['certificate-authority']).forEach(cluster => {
  cluster.cluster['certificate-authoriy'] = path.resolve(cfgPath, cluster.cluster['certificate-authority']);
});

config.users.filter(user => user.user['client-certificate']).forEach(user => {
  user.user['client-certificate'] = path.resolve(cfgPath, user.user['client-certificate']);
});

config.users.filter(user => user.user['client-key']).forEach(user => {
  user.user['client-key'] = path.resolve(cfgPath, user.user['client-key']);
});
```

raszi: Well, that is why I was asking which solution you prefer. This was the immutable solution, the mutable looked something similar to your approach without the duplication.

raszi: I did not know `path.resolve()`. Thanks!

silasbw:
Published changes as version `3.16.1` -- thanks again @raszi!

Issue #129, opened 2017-07-12
Use swagger specs for generating object chains


silasbw:
The [Python client](https://github.com/kubernetes-incubator/client-python) does something similar by leverage code generated by swagger-codegen.

kubernetes-client could use [swagger-js](https://github.com/swagger-api/swagger-js) to generate code dynamically using swagger specs distributed with kubernetes-client or using swagger specs hosted on the kube-apiserver.

Pupix:
Does it really have to be this verbose?

```js
const kubernetes = require('kubernetes-client');
const spec = require('./swagger-1.8.json');
const http = new kubernetes.Request(kubernetes.config.fromKubeconfig());
const client = new kubernetes.SwaggerClient({ spec, http });
```

Using the above example, I should somehow find the `swagger.json` file for my kubernetes version, then initialize the Request and finally create the `SwaggerClient`.

Is it not possible for the module to come with predefined swagger files for X kubernetes versions? Or can't I just specify the kubernetes version I am using? Or can't it find the correct swagger file from my server?

Cheers,
Pupix

silasbw:
Consider this PR the "primitives" we'd use to swagger support: the API contract of the generated client and the implementation of the code that generates it.  So, I think I agree with what you're suggesting:

* No, it doesn't need to be this verbose :smile: 
* Yes we should ship with predefined clients; and
* yes make async client creation easy by pointing at the kube-apiserver swagger.json.

If you have time to implement a "higher-level" API for the swagger stuff, we'd :heart: that.  It would be useful to have an example of what that API contract might be (and discuss it); and also to have a concrete implementation to work from (even if it wasn't perfect).

silasbw:
This is complete.

PR #162, opened 2017-12-30
feat(Swagger): experimental support for Swagger.

silasbw:
Converts a Swagger spec into a kubernetes-client API client.

https://github.com/godaddy/kubernetes-client/issues/129

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -120,7 +120,9 @@ class Component {
       //
       if (parameter) {
         if (!parent.parameter) {
-          throw new Error('Created Component, but require templated one.');
+          throw new Error('Created Component, but require templated one. ' +
+                          'This is a bug. Please report: ' +
+                          'https://github.com/godaddy/kubernetes-client/issues');
         }
         parent.templatedEndpoints.push(new Endpoint({
           name: endpoint.name,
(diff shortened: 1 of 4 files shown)


lib/swagger-client.js
```diff
@@ -0,0 +1,263 @@
...
+ *   /api/v1/namespaces/{namespace}/pods -> api.v1.namespaces(namespace).pods
+ *
+ * Iterate over a Paths Object [3] to generate whole API client.
+ *
+ * [1]: https://swagger.io/specification/#pathItemObject
+ * [2]: https://swagger.io/specification/#pathTemplating
+ * [3]: https://swagger.io/specification/#pathsObject
+ */
+
+const assign = require('assign-deep');
```
gergelyke: maybe using a more popular solution, like [lodash.assign](https://www.npmjs.com/package/lodash.assign)?

lib/swagger-client.js
```diff
@@ -0,0 +1,263 @@
...
+      }
+      parent = parent[split];
+
+      //
+      // Path Template parameter: save it and walk it once the user specifies
+      // the value.
+      //
+      if (parameter) {
+        if (!parent.parameter) {
+          throw new Error('Created Component, but require templated one.');
```
gergelyke: can we make it a bit more descriptive? like as a developer, what should I do to fix this?

PR #169, opened 2018-01-05
Add NamespaceList v1 core enumeration

christianjoun:
This was achieved by adding the string 'namespaces' to groupResources in core.js. Thus Accessing NamespaceList v1 core api
can be easily done using
`core.namespaces.get(print);`.
To maintain previous functionality I renamed the default namespace (directly on the API) to `namespace`
(single instead of plural) and updated the alias ns.  Thus accessing deployments on a given NAMESPACE can be done using
`ext.namespace(NAMESPACE).deployments('').get(print);`

### DIFF
--- a/lib/api-group.js
+++ b/lib/api-group.js
@@ -44,7 +44,7 @@ class ApiGroup {
     //
     // Create the default namespace so we have it directly on the API
     //
-    this.namespaces = new Namespaces({
+    this.namespace = new Namespaces({
       api: this,
       parentPath: this.path,
       namespace: options.namespace,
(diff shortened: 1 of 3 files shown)


silasbw:
@christianjoun, thanks for this and sorry it slipped through the review cracks :/.

This would break existing users.  Also, "`core.namespace.get(print)`" should already work (and if it doesn't that's a bug).  Can you verify if `core.namespace.get(print)` has the expected result or not?

Ewocker:
Yes, it does work in 4.0.0 now. @christianjoun works with me on that. Thanks for the reply.

PR #173, opened 2018-01-08
chore(package-lock): Recreate package-lock file to exclude artifactory URIs.

gisenberg:


gisenberg:
Over-the-shoulder signoff from @silasbw

PR #183, opened 2018-02-07
Promise actions typings

amiram:
Add types for the promised functions: getPromise, postPromise, putPromise, deletePromise, patchPromise


silasbw:
Thanks @amiram!

@gisenberg Can you help review this?

silasbw:
@amiram, did you notice https://github.com/godaddy/kubernetes-client#experimental-support-for-promises-and-asyncawait?  Or do you have an application where you want to call both the callback-based and promised-based API?  I don't think we expected that, but if there are some use cases for it we can probably add a nice way to do it.

amiram:
I saw the experimental support for promises. I did the PR according to current implementation.
Currently if you use the regular `get` function to get the promise you get an `any`, which is not exactly accurate since you get a stream or promise if you don't pass a callback, and void if you do. Usually you split the function to two versions, one with a callback and another without, but then you have to use casting to cast to a promise (because it might be a stream). Since I wanted more clean code I used the getPromise function, and for this I made the PR.
Anyway, this kind of implementation of stream or promise is not so common. I saw some discussion about removing the stream support in a major version. If you do that, typings can be changed to:
`get(callback: (err: any, value: any) => void): void`
`get(options: ApiRequestOptions, callback: (err: any, value: any) => void): void`
`get(options?: ApiRequestOptions): Promise<any>`

silasbw:
Thanks for helping with getting typings improved.  If you have issues with the current implementation / auto-generated ones, let us know.

PR #193, opened 2018-02-11
feat(promises): if caller doesn't specify a callback, return a promise

### LINKED ISSUE #189: Request with no callback results in error
The following request will cause an error, because there is no callback defined:
```javascript
const Api = require('kubernetes-client');
const core = new Api.Core(Api.config.getInCluster());
core.namespaces(process.env.POD_NAMESPACE).pod(process.env.POD_NAME).patch({
  body: {
    metadata: {
      labels: {
        app: 'working'
      }
    }
  }
})
```
```javascript
/usr/src/app/node_modules/kubernetes-client/lib/base.js:15
    cb(null, result.body);
    ^

TypeError: cb is not a function
    at /usr/src/app/node_modules/kubernetes-client/lib/base.js:15:5
    at Request.request [as _callback] (/usr/src/app/node_modules/kubernetes-client/lib/request.js:65:7)
    at Request.self.callback (/usr/src/app/node_modules/request/request.js:186:22)
    at emitTwo (events.js:126:13)
    at Request.emit (events.js:214:7)
    at Request.<anonymous> (/usr/src/app/node_modules/request/request.js:1163:10)
    at emitOne (events.js:116:13)
    at Request.emit (events.js:211:7)
    at IncomingMessage.<anonymous> (/usr/src/app/node_modules/request/request.js:1085:12)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:111:20)
    at IncomingMessage.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1055:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

Would it be possible to just add a check [here](https://github.com/godaddy/kubernetes-client/blob/f8f1d356244055b6bdebe5587c3142d5f69f3ceb/lib/base.js#L15) to see if a callback exists?

silasbw:
Resolves #189

BREAKING CHANGE: This removes the (deprecated) `request`-like behavior of
returning a stream when the caller omits a callback. Use `.getStream` (instead
of `.get` without a callback) to get a stream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,22 +71,9 @@ const core = new Api.Core(Api.config.fromKubeconfig());
 
 ### **Experimental** support for promises and async/await
 
-kubernetes-client exposes **experimental** support for promises via
-the `promises` option passed to API group constructors. The API is the
-same, except for the functions that previously took a callback
-(*e.g.*, `.get`). Those functions now return a promise.
-
-```js
-// Notice the promises: true
-const core = new Api.Core({
-  url: 'http://my-k8s-api-server.com',
-  version: 'v1',  // Defaults to 'v1'
-  promises: true,  // Enable promises
-  namespace: 'my-project' // Defaults to 'default'
-});
-```
-
-and then:
+kubernetes-client has **experimental** support for promises. If you
+omit callbacks an HTTP method function (*e.g.*, `.get`), it will
+return a promise.
 
 ```js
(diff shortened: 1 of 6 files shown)


silasbw:
Depends on https://github.com/godaddy/kubernetes-client/pull/192

PR #203, opened 2018-02-26
Fix request options parsing in ApiGroup

johngiudici:
The readme indicates that bare request options can be passed through when instantiating ApiGroup objects. It looks like there was a regression introduced with https://github.com/godaddy/kubernetes-client/pull/161. I've included a fix here, along with tests for all of the parameters defined in the docs.


silasbw:
Doh, thanks @johngiudici.  I'll push this fix out to npm in the next hour or two.

silasbw:
...published as 4.0.1.

Thanks again!

PR #207, opened 2018-03-07
feat(dynamic clients): introduce more user-friendly Client classes

silasbw:
Client and SyncClient abstract away the HTTP request adapter.


lib/client.js
```diff
@@ -0,0 +1,20 @@
...
+class Client extends SwaggerClient {
+  constructor(options) {
+    const http = new Request(options.config);
+    return new Promise((resolve, reject) => {
+      http.request('GET', { path: '/swagger.json' }, (err, res) => {
+        if (err) return reject(err);
+        if (res.statusCode !== 200) {
+          return reject(new Error(`Failed to get swagger.json: ${ res.statusCode }`));
+        }
+        resolve(super({ http, spec: res.body }));
```
silasbw: This is going to cause problems for Node < 6.  Should we drop support for Node 4 and Node 5?

bmeck: Node 4 EOLs at the end of April : https://github.com/nodejs/Release , seems fine to not go out of our way at this point? It is slightly sad that we won't hit the EOL date, but don't see much sense in going out of our way for this.

examples/sync-client-version.js
```diff
@@ -0,0 +1,20 @@
+//
+// Create an API client based on a specified API version. kubernetes-client uses
+// included swagger specification files to implement this.
```
bmeck: can we remove the "to implement this" bit, confused me on first read.

lib/index.js
```diff
@@ -1,16 +1,24 @@
+//
+// Deprecated interface
+//
```
bmeck: do we have any documentation on this being deprecated? might want to add that too. Node generally does documentation -> runtime warning -> removal over time for their deprecation policy. IDK if we want to follow, but docs seem good either way.

silasbw: :+1: Sounds good.  I'm going to rewrite README.md next, so I'll start your suggested deprecation process with that re-write.

PR #213, opened 2018-03-08
refactor(swagger-client): add `._addSpec` logic

silasbw:
This change facilitates support for CRDs.

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -113,7 +113,8 @@ class Component {
     //
     Object.keys(spec.paths)
       .map(name => {
-        const splits = name.replace(/(^\/)|(\/$)/g, '').split('/');
+        const leadingAndTrailingSlashes = /(^\/)|(\/$)/g;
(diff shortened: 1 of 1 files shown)


lib/swagger-client.js
```diff
@@ -93,6 +93,36 @@ class Component {
...
+    // E.g., add /api/v1/namepaces/{namespace} before adding /api/v1/namepaces
+    //
+    // This is important because ._addEndpoint constructs Component objects on
+    // demand, and Component requires specifying if it's templated or not. If we
+    // cause ._addEndpoint to construct a un-templated Component, templated
+    // operations that share the Components subpath will not work.
+    //
+    Object.keys(spec.paths)
+      .map(name => {
+        const splits = name.replace(/(^\/)|(\/$)/g, '').split('/');
```
gergelyke: can you add a quick comment on what this regex does? or maybe better:

```
const thePurposeOfThisRegex = /(^\/)|(\/$)/g;
const splits = name.replace(thePurposeOfThisRegex, '').split('/');
```

PR #222, opened 2018-03-10
docs(README.md): Re-write README.md for the Swagger-based client

silasbw:


examples/basic.js
```diff
@@ -0,0 +1,41 @@
...
+    //
+    // Fetch the Deployment we just created.
+    //
+    const deployment = await client.apis.apps.v1.namespaces('default').deployments(deploymentManifest.metadata.name).get();
+    console.log('Deployment: ', deployment);
+
+    //
+    // Remove the Deployment we created.
+    //
+    const removed = await await client.apis.apps.v1.namespaces('default').deployments(deploymentManifest.metadata.name).delete();
```
gisenberg: This looks like an accidental syntax error.

silasbw: :crying_cat_face:

examples/basic.js
```diff
@@ -0,0 +1,41 @@
+//
+// Demonstrate some of the basics.
+//
+const Client = require('kubernetes-client').SyncClient
```
gisenberg: nit: missing semicolon.

examples/basic.js
```diff
@@ -0,0 +1,41 @@
+//
```
gisenberg: I found this file confusing because of the usage of `SyncClient` - based on the naming, I expected to receive a client that had only synchronous methods, but on further investigation, the constructor is synchronous and the methods hanging off of it are async.

I'm not sure which is more intuitive.

silasbw: It's not you, it's me :)  addressed in https://github.com/godaddy/kubernetes-client/pull/225

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+const core = new Api.Core({
+  url: 'http://my-k8s-api-server.com',
+  version: 'v1'  // Defaults to 'v1'
+});
+
+function print(err, result) {
+  console.log(JSON.stringify(err || result, null, 2));
+}
+
+core.namespaces('default').replicationcontrollers('http-rc').get(print);
```
gisenberg: This uses the `default` namespace, but the description of the example on line 32 says it is going to use the `my-project` namespace - is this intentional?

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+
+```js
+const Api = require('kubernetes-client');
+const core = new Api.Core(Api.config.fromKubeconfig());
+```
+
+### **Experimental** support for promises and async/await
+
+kubernetes-client has **experimental** support for promises. If you
+omit callbacks an HTTP method function (*e.g.*, `.get`), it will
```
gisenberg: nit: comma after "omit callbacks"

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+
+```js
+core.namespaces('default').replicationcontrollers('http-rc').get()
+  .then(result => print(null, result));
+```
+
+or with `async/await`:
+
+```js
+print(null, await core.namespaces('default').replicationcontrollers('http-rc').get());
```
gisenberg: This calls attention to `print` being a little weird - maybe just console.log the result?

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+### Using the correct API group and version
+
+kubernetes-client client includes functionality to help determine the
+correct Kubernetes API group and version to use based on manifests:
+
+```js
+const Api = require('kubernetes-client');
+const api = new Api.Api({ url: 'http://my-k8s-api-server.com' });
```
gisenberg: Does this code get easier to read if line 115 turns into `const Api = require('kubernetes-client').Api;``?

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+ext.thirdpartyresources.post({ body: newResource }, print);
+```
+
+and then extend an `ThirdPartyResource` API client with your new resources:
+
+```js
+const thirdPartyResources = new Api.ThirdPartyResources({
+  url: 'http://my-k8s-api-server.com',
+  group: 'kubernetes-client.io',
+  resources: ['customresources']  // Notice pluralization!
```
gisenberg: nit: `customresources` => `customResources`

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+});
+
+// Access `customresources` as if they were a regular Kubernetes object
+thirdPartyResources.ns('default').customresources.get(print);
+thirdPartyResources.addResource('newresources');  // Notice pluralization!
```
gisenberg: nit: `newresources` => `newResources`

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+
+You can access logs in a similar fashion:
+```js
+const stream = core.ns('default').po('http-123').log.getStream({ qs: { follow: true } });
+stream.on('data', chunk => {
+  process.stdout.write(chunk.toString());
+});
+```
+
+**Note:** the kube-apiserver will close watch connections eventually
```
gisenberg: I found this difficult to parse. Maybe something like this?

> the kube-apiserver eventually closes `watch` connections

README-PRE-5.md
silasbw: the lower case is correct for the (deprecated) third part resources.

README-PRE-5.md
```diff
@@ -0,0 +1,374 @@
...
+according to the
+[`--min-request-timeout`](http://kubernetes.io/docs/admin/kube-apiserver/
+command line argument. kubernetes-client does not attempt to reconnect
+when the kube-apiserver closes a connection.
+
+### Authenticating
```
gisenberg: Authenticating or Authentication?

README-PRE-5.md
gisenberg: :(

PR #223, opened 2018-03-12
docs(generate): use swagger-js-codegen to generate markdown docs

silasbw:
Include mustache templates and markdown for the included swagger specs.


package.json
```diff
@@ -7,8 +7,9 @@
   "scripts": {
     "check-coverage": "istanbul check-coverage",
     "coverage": "istanbul cover ./node_modules/.bin/_mocha && istanbul report cobertura",
+    "docs": "for v in 1.7 1.8 1.9; do scripts/docs.js --spec lib/specs/swagger-$v.json.gz --output docs/$v.md; done",
```
gisenberg: This won't work cross-platform.

jcrugzz:
@silasbw these should probably be in `.npmignore`

package.json
jeffpearce: For my own learning - which part will break? I'm assuming on Windows?

scripts/docs.js
```diff
@@ -0,0 +1,100 @@
...
+   */
+  alphanumeric: function () {
+    return function (text, render) {
+      return render(text).replace(/\W/g, '');
+    };
+  },
+
+  /**
+   * Replace newline characters in rendered text with <br/>s.
+   * @returns {funtion} function
```
jeffpearce: nit: `{function}`

scripts/docs.js
```diff
@@ -0,0 +1,100 @@
...
+   */
+  markdownBreaks: function () {
+    return function (text, render) {
+      return render(text).replace(/\n/g, '<br/>');
```
jeffpearce: Fo we need to support CRLF for Windows?

gisenberg: It might be worth calling out these line endings specifically in `.gitattributes`

scripts/docs.js
```diff
@@ -0,0 +1,100 @@
...
+    };
+  },
+
+  /**
+   * When in a method section, return the full kubernetes-client name.
+   * @returns {funtion} function
```
jeffpearce: nit: `{function}` again

package.json
gisenberg: 1. bash-style for-loop syntax won't work on Windows (syntax error)
1. executing scripts/docs.js directly won't work on Windows (not an executable)

silasbw:
@jcrugzz, we're declaring the directories / files that need to be included (instead of .npmignoring things):

https://github.com/godaddy/kubernetes-client/blob/master/package.json#L28

PR #251, opened 2018-04-07
Automatically load kubeconfig when its provided as a path to the file

ywangii:
Talked with @jcrugzz, we probably need a back patch for this fix (probably for version 4?) since `trimmer` is still using 3.x.x version.


silasbw:
I don't think you'll need to backport.  5.X.X still supports the old API.

Issue #220, opened 2018-03-10
Add config examples for constructing clients


silasbw:
All the current examples rely on `config.fromKubeconfig()`.  Add examples that manually specify some of the config options.

qrpike:
AGREED!

idirouhab:
I added the basic auth, is there any other proposal?

silasbw:
Basic auth works.  It's documented in the README.md.  Does that work for you @idirouhab ?

This issue is about adding an example to `examples/`

idirouhab:
it's works, I was just asking to contribute and create them, just wanted to know which examples do you need

silasbw:
We'd love it if you wanted to contribute the examples.  I edited the original issue comment to provide details on what I think would be a good example.

Sorry I overlooked that you added the README.md documentation.  Thank you for doing that :heart:

cmosgh:
Any chance of adding a config for an app running in a pod?

flowstate:
I'm new to Node and this client. I have had to dig through the config code for a couple of hours to get as far as I have, which is to get rejected as 'system:anonymous'. Now I have no idea how to continue, as I've never had this issue with any k8s cluster I've ever made.

I point this out here because I feel like the client config section really needs love. I'm super close to just giving up on this project because I don't have enough of a sense of how it does anything (like deciding what user I am, for instance) to reason about how to fix it myself.

I understand exactly how hard open source is, and I'm grateful for the repo in general, but I mention all this to underscore just how important this area of the readme is.

I hope it is explained further in the future. Maybe a list of the expectations / requirements for successful config loading?

cmosgh:
What I digged out of the docs is something pretty neat
```
import kubernetesClient from "kubernetes-client";

const Client = kubernetesClient.Client;
const config = kubernetesClient.config;

const client = new Client({
  config: onCluster ? config.getInCluster() : config.fromKubeconfig(),
  version: "1.9"//there is a list in docs of the options you have
});
```
From there on you can use all the apis defined on the client

cmosgh:
The idea is that you'll deploy this in a node container and of course you have to create a service account and a cluster role binding so the pod can run with the proper rights.

silasbw:
@flowstate I agree, documentation and code examples could be much clearer.  Can you share some code or explain more what you're trying to do and where code is running in relation to your kubernetes cluster?

PR #263, opened 2018-05-08
Adding create client example using basic-auth

idirouhab:
Fixing #220


idirouhab:
Actually me neither, I even changed the way to get the properties to avoid issues.

PR #270, opened 2018-05-23
Add getInCluster documentation example.

### LINKED ISSUE #269: Provide example of using in pod authentication using getInCluster
There isn't an example in the readme of using `getInCluster()` for authenticating to the API from within a kubernetes pod.  It appears that the code to do this exists [here](https://github.com/godaddy/kubernetes-client/blob/5.3.0/lib/config.js#L17-L45).

I'll be sending a pull request to address this soon.

godber:
Adds example of using `getInCluster()` for config, closes #269

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,14 @@ const client = new Client({
 })
 ```
 
+or from within a Pod:
+
+```js
+const Client = require('kubernetes-client').Client
+const config = require('kubernetes-client').config;
+const client = new Client({ config: config.getInCluster() });
+```
+
 ## Basic usage
 
 kubernetes-client translates Path Item Objects \[[1]\] (*e.g*.,


godber:
I think this is the right thing to do, I'll be using this soon and update if it's not quite right.

README.md
```diff
@@ -64,6 +64,14 @@ const client = new Client({
 })
 ```
 
+or from within a Pod:
+
+```js
+const Client = require('kubernetes-client').Client
+const config = require('kubernetes-client').config;
+const client = new Client({ config: config.getInCluster() });
+```
```
silasbw: you need to specify a version or, probably better, add a:

```js
await client.loadSpec();
```

at the end of this example.

README.md
```diff
@@ -64,6 +64,14 @@ const client = new Client({
 })
 ```
 
+or from within a Pod:
+
+```js
+const Client = require('kubernetes-client').Client
```
silasbw: meh, missing a `;`.

(FWIW, we're moving the code base to standard.js asap, but for now, please add the `;`)

godber:
Happy to sort these out.  I added the other missing `;` and the call to `loadSpec()`.  I still haven't had a chance to run my code in the pod and verify that `getInCluster()` works as advertised, but it looks good!

PR #275, opened 2018-06-05
Delete resource bugfix

NoChance777:
When trying to delete resource with ...delete() it causes exception "Cannot read property 'qs' of undefined".

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -91,6 +91,8 @@ class BaseObject extends CallableObject {
       options = {};
     } else if (typeof options === 'string') {
       options = { name: options };
+    } else if (typeof options === 'undefined') {
+      options = {};
     }
     this.api.delete({ path: this._path(options), qs: options.qs, body: options.body },
       cb200(cb));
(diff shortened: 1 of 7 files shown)


NoChance777:
HI there! I don't know why checks were failed

lib/base.js
```diff
@@ -91,6 +91,8 @@ class BaseObject extends CallableObject {
       options = {};
     } else if (typeof options === 'string') {
       options = { name: options };
+    } else if (typeof options === 'undefined') {
```
silasbw: How are you calling delete?  Are you also leaving `cb` undefined?  If so, won't `this.api.delete` fail when it invokes the cb?

NoChance777: Hi! I've been using this lib since version 3 and only update version, but not a code. 
Here is example of using
```javascript
const api = new Api.Api(k8sConfig); //get config from file
this.api = api.group(apiVersion);  //use batch/v1 for jobs
...
await this.api.ns(defaultNS).jobs(jobName).delete();      
```

NoChance777: passing empty object into delete method works well, but it isn't documented

silasbw:
Ok, sounds good.  Can you remove your changes to `package-lock.json`?  I think this is good to go after that.

NoChance777:
done. but keep in mind that with original package-lock.json tests will be failed

Issue #204, opened 2018-02-28
Support for 'rollout undo' deployment


daseyffert:
I am curious if there is any feature with this client that takes advantage of kubectl rollout undo command?

If there isn't, any guidance on being able to implement this feature would be very helpful as well.

I am dealing with Kubernetes version 1.7 and 1.8.

silasbw:
I'm not sure how that's implemented, but in theory it should be implementable using the Kubernetes API via kubernetes-client.  This isn't an issue with kubernetes-client specifically, but it would be useful to include this as an example.

If you end up implementing this, can you share the code with us so we can include it as an example?

daseyffert:
I wasn't able to figure it out using kubernetes-client so I ended up implementing a different way.

In order to roll back you need to run a POST command to the url path: 
/apis/extensions/v1beta1/namespaces/<! NAMESPACE>/deployments/<! DEPLOYMENY>/rollback

The body of the POST command looks as follows:
var postBody =
        {
            "kind": "DeploymentRollback",
            "apiVersion": "extensions/v1beta1",
            "name": <! NAME_OF_YOUR_DEPLOYMENT>,
            "rollbackTo":
                {
                    "revision": <! REVISION_NUM_YOU_WANT_TO_ROLLBACK_TO>
                }
        };

Hope this is useful.

PR #302, opened 2018-07-27
feat(example): example for creating, updating, and rolling back a deployment

tkporter:
Example in response to issue #204 
This creates a deployment, patches it with a new version of an image, and rolls back to the original image.

### DIFF
--- a/examples/deployment-create-patch-rollback.js
+++ b/examples/deployment-create-patch-rollback.js
@@ -32,6 +32,7 @@ async function main() {
     });
     console.log('Update: ', updateImage);
 
+    // Rollback to nginx:1.7.9
     const rollback = await client.apis.apps.v1beta1.namespaces('default').deployments('nginx-deployment').rollback.post({
       body: {
         kind: 'DeploymentRollback',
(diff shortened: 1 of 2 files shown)


examples/deployment-create-patch-rollback.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+      }
+    });
+    console.log('Update: ', updateImage);
+
+    // Rollback to nginx:1.7.9
+    const rollback = await client.apis.apps.v1beta1.namespaces('default').deployments('nginx-deployment').rollback.post({
+      body: {
+        kind: 'DeploymentRollback',
+        apiVersion: 'apps/v1beta1',
```
silasbw: Can you use `apps/v1` here?

tkporter: After testing this out, v1.10 only supports rollbacks for apiVersion 'apps/v1beta1'.
```
Error:  { Error: DeploymentRollback in version "v1" cannot be handled as a DeploymentRollback: no kind "DeploymentRollback" is registered for version "apps/v1"
```

PR #308, opened 2018-08-08
feat(browser): add support for browser-request in request class, resolves #234

### LINKED ISSUE #234: Support `fetch` as an `http` backend for the swagger client
The swagger client currently depends on a [request](https://www.npmjs.com/package/request)-based [HTTP request object](https://github.com/godaddy/kubernetes-client/blob/master/lib/request.js) for making HTTP calls.  This is a blocker for running kubernetes-client on browsers.

Implement a `fetch`-based HTTP request object.

cfellin1:


lib/request.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
-const request = require('request');
+// If in browser, use request-browser; else use normal request.
+const request = Window ? require('request-browser') : require('request');
```
silasbw: `typeof Window !== 'undefined' ? ...`

Also, where does request-browser come from?  do we need to update package.json and package-lock.json?

xmlking:
can we use  to `fetch` for browser and use `node-fetch` as polyfill  for node ?

cfellin1:
@xmlking What's the benefit to justify changing the request wrapper class?

xmlking:
Fetch is natively promise. No polyfills required for browsers.

jeffpearce:
So it's polyfill for node vs polyfill for browser? Did I understand correctly? What are the tradeoffs of one vs the other?

xmlking:
Yes. Backend polyfill can be removed when node supports it. Keeping browser more stable is important in my opinion

cfellin1:
Closing this PR: if we go with the fetch route, the end result will look pretty different than this.

PR #322, opened 2018-09-04
fix(kube-config): use KUBECONFIG env var if set

JacopoDaeli:
**NOTE**
Do we eventually want to support multiple config files in the future?
https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#create-a-second-configuration-file


silasbw:
Should we go straight to supporting the merge?  What do we get by supporting the first kubeconfig file in the path?  That seems like it's an unexpected thing to support.

JacopoDaeli:
I think have one single `kubeconfig` file is the most common case, and with this PR covers it. I see it like a mid step. However, I am ok with supporting the merge right away. Whatever u folks feel better. I have no strong opinion on this.

silasbw:
OK, sounds good to me.

PR #326, opened 2018-09-06
feat(config): support multiple config files in KUBECONFIG

JacopoDaeli:
Adding support for multiple config files in `KUBECONFIG` env variable.
Official ref: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#create-a-second-configuration-file

Output of `console.log(yaml.dump(merge.all(configs)))` ([ref to configs](https://github.com/godaddy/kubernetes-client/pull/326/files#diff-f7bf2b665273dfa66ffa4ad7c0a52bb9R183)) is:
image
which matches the output of `kubectl config view`:
image

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -10,7 +10,7 @@ const caPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt';
 const tokenPath = '/var/run/secrets/kubernetes.io/serviceaccount/token';
 const namespacePath = '/var/run/secrets/kubernetes.io/serviceaccount/namespace';
 
-function defaultConfigPath() {
+function defaultConfigPaths() {
   if (process.env.KUBECONFIG) {
     // From https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable
     // KUBECONFIG can support multiple config files.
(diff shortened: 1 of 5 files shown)


JacopoDaeli:
Follow up after discussion in https://github.com/godaddy/kubernetes-client/pull/322

JacopoDaeli:
Of course. I will deal with it ASAP.

lib/config.js
```diff
@@ -159,27 +158,34 @@ module.exports.fromKubeconfig = fromKubeconfig;
...
+      user.user['client-key'] = path.resolve(configDir, user.user['client-key']);
+    });
+  }
 
   return config;
 }
 
 function loadKubeconfig(cfgPath) {
-  cfgPath = cfgPath || defaultConfigPath();
+  const cfgPaths = cfgPath ? [cfgPath] : defaultConfigPath();
```
jcrugzz: @JacopoDaeli was thinking, should we support passing an array of config paths in here as well? i.e. 

```js
let cfgPaths;
if (!cfgPath) { cfgPaths = defaultConfigPath(); }
else if(Array.isArray(cfgPath)) { cfgPaths = cfgPath; }
else { cfgPaths = [cfgPath]; }
```

JacopoDaeli: I was thinking that too. Not sure. In any case, that doesn't hurt, but is it useful?
@silasbw what do you think?

PR #329, opened 2018-09-11
WIP: Implement initial handling for POST /pods/:pod/exec 

gempesaw:
This doesn't work at all yet - I'm just opening the PR early for fun. I'm currently stuck trying to figure out how to pass the ca, cert, & key for the WS connection. Whenever I finish, I'm planning to rebase to fit y'all's commit message format.


silasbw:
Cool!  Maybe @3rd-Eden can give some advice on using client certs with `ws`.

gempesaw:
Cool, it turns out I had been experimenting without `stdin=true&stderr=true&stdout=true` query parameters in the exec to Kubernetes; apparently those are mandatory. This is now in a working state, and I also rebased onto master to pick up changes in `package-lock.json` since last week.

I've added a few questions inline! Cheers.

lib/request.js
```diff
@@ -23,6 +24,100 @@ function refreshAuth(type, config) {
...
 
+const execChannels = [
+  'stdin',
+  'stdout',
+  'stderr',
+  'error',
+  'resize'
+];
+
+/**
```
gempesaw: I'm not in the habit of writing JSDoc, so I'm happy to take any feedback here.

lib/request.js
```diff
@@ -23,6 +24,100 @@ function refreshAuth(type, config) {
...
+  if (!command || !Array.isArray(command)) {
+    return cb(new Error(`expected command to be an array of strings, not: ${JSON.stringify(command)}`));
+  }
+
+  // The command query parameter gets specified multiple times to
+  // indicate arguments to the initial binary
+  // (command=ls&command=-al), so it cannot be handled in an object
+  // like the rest of the qs.
+  const commandAsParams = command.map((cmd) => `command=${encodeURIComponent(cmd)}`);
+  delete qs.command;
```
gempesaw: I'd prefer not to mutate here if possible, but I assume something like 

```js
const { qs: { command, ...rest } } = options;
```

is out of the question? :P

lib/request.js
```diff
@@ -23,6 +24,100 @@ function refreshAuth(type, config) {
...
+    const message = Buffer.from(msg.slice(1), 'base64').toString('ascii');
+    messages.push({ channel, message });
+  });
+
+  ws.on('error', (err) => {
+    err.messages = messages;
+    cb(err, messages);
+  });
+
+  ws.on('close', (code, reason) => cb(null, {
```
gempesaw: I wasn't sure what the response format for this should be. The joined command output is available on `res.body`, but I figure it may also be useful to see the messages, websocket exit code, and reason (if any).

silasbw: yeah, i'm not sure either.  I like smaller incremental PRs/changes.  So one thing we could do is try what you have here, but don't advertise support yet, and expect to iterate on the exact result format, etc.

lib/request.js
silasbw: I think the `qs` suggestion about might help resolve this.

lib/request.js
```diff
@@ -23,6 +24,100 @@ function refreshAuth(type, config) {
...
+/**
+ * Determine whether a failed Kubernetes API response is asking for an upgrade
+ * @param {object} body - response body object from Kubernetes
+ * @property {string} status - request status
+ * @property {number} code - previous request's response code
+ * @property {message} message - previous request response message
+ * @returns {boolean} Upgrade the request
+ */
+
+function isUpgradeRequired(body) {
```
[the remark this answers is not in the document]
gempesaw: Unfortunately, the response to making a POST to `/pods/:pod/exec` without the required websocket headers doesn't give much to go on:

```
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< Date: Mon, 17 Sep 2018 18:40:02 GMT
< Content-Length: 139
< 
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Upgrade request required","reason":"BadRequest","code":400}
```

lib/request.js
```diff
@@ -23,6 +24,100 @@ function refreshAuth(type, config) {
...
+/**
+ * Construct the query parameters necessary for performing an execAction
+ * @param {object} qs - query parameter object as passed by user
+ * @property {string[]} command - (required) command and arguments to be executed
+ * @property {string} container - (optional) container in which to execute the command; optional if there's only one container in the pod
+ * @property {boolean} tty - (optional) whether to hold the websocket connection open for an interactive session (not yet implemented)
+ * @param {function} cb - callback to handle the response
+ * @returns {string} All necessary query parameters as a string ready for inclusion in URL
+ */
+function buildExecQueryParams(qs, cb) {
```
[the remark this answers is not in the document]
gempesaw: Oh, yes - that should make this much cleaner.

lib/request.js
```diff
@@ -23,6 +24,100 @@ function refreshAuth(type, config) {
...
+  }
+
+  // The command query parameter gets specified multiple times to
+  // indicate arguments to the initial binary
+  // (command=ls&command=-al), so it cannot be handled in an object
+  // like the rest of the qs.
+  const commandAsParams = command.map((cmd) => `command=${encodeURIComponent(cmd)}`);
+  delete qs.command;
+
+  const execParams = Object.assign(qs, {
```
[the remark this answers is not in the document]
gempesaw: Hmm, so the caller must know to include them in the body on every call? aka usage example would change to

```node
client.api.v1.namespaces('namespace_name').pods('pod_name').exec.post({
  qs: {
    command: [ 'ls', '-al' ],
    stdin: true, // hmm perhaps stdin isn't necessary, i'll have to double check
    stdout: true,
    stderr: true
  }
})
```

Unfortunately, the error response while I was forgetting to include `stdout/stderr` query params was just a 400 Error with no message, so it may be a bit difficult for users to debug... but if it's in the `examples/` usage, hopefully that's enough guidance? 

Happy to make the change, and agree it makes sense since kubernetes-client isn't in the habit of adding query parameters anywhere else, just wondering about troubleshooting :P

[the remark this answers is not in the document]
silasbw: Yeah, I'm suggesting that the user needs to include those on every call for the reason you mention about eschewing adding query parameters. I think the example/ you wrote will help.  I think is there's enough interest and/or common patterns it could make sense to have a separate module focused on helping exec things (e.g., `kubernetes-exec`).

lib/request.js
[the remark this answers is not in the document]
silasbw: what...ok, thanks for checking :+1:

lib/request.js
```diff
@@ -23,6 +25,63 @@ function refreshAuth(type, config) {
...
+
+/**
+ * Upgrade a request into a Websocket transaction & process the result
+ * @param {ApiRequestOptions} options - Options object
+ * @param {callback} cb - The callback that handles the response
+ */
+
+function upgradeRequest(options, cb) {
+  const queryParams = qs.stringify(options.qs, { indices: false });
+    const wsUrl = `${options.baseUrl}/${options.uri}?${queryParams}`;
```
silasbw: spacing nit

PR #332, opened 2018-09-20
feat(auth): support auth through iam-authenticator, add example, Closes #306

### LINKED ISSUE #306: Please support AWS Authenticator for EKS
Amazon EKS requires the user generates a token through the AWS Authenticator, using the `exec` option on the user property in the kubeconfig file. kubernetes-client does not appear to support this option, so we cannot use it to talk to our EKS clusters.

https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html

cfellin1:


examples/iam-auth.js
```diff
@@ -0,0 +1,50 @@
+/* eslint no-console:0 */
+//
+// Create an API client for an AWS EKS cluster using IAM authentication
+// Note: In order for this to work, you must set the environment
+// variable AWS_PROFILE
```
silasbw: `AWS_PROFILE` is used here: https://github.com/godaddy/kubernetes-client/pull/332/files#diff-7b7c8857ce22c1739c6ad1ea186ca9ddR22

Should this comment be in that file?

examples/eks-auth.js
```diff
@@ -0,0 +1,51 @@
+/* eslint no-console:0 */
+//
+// Create an API client for an AWS EKS cluster
```
silasbw: What's the difference between this example and the one below?

examples/iam-auth.js
```diff
@@ -0,0 +1,53 @@
+/* eslint no-console:0 */
```
silasbw: Can you add a link to this example from the README.md?

PR #333, opened 2018-09-20
feat(config): read exec arguments from kubeconfig

cfellin1:


lib/config.js
```diff
@@ -141,6 +141,25 @@ function fromKubeconfig(kubeconfig, current) {
       };
     }
 
+    if (user.exec) {
+      const env = {};
+      user.exec.env.forEach((kv) => {
```
silasbw: naming nit: `(kv) =>` to `variable =>`

PR #335, opened 2018-09-25
Fix missing addCustomResourceDefinition in type definitions, add support for /status and /scale CRD endpoints

### LINKED ISSUE #334: Method addCustomResourceDefinition is not resolved in TypeScript
The following code:

```typescript
const Client = Api.Client1_10;
const config = Api.config;
const client = new Client({ config: config.fromKubeconfig() });

client.addCustomResourceDefinition(crdSchema);
```

Fails to compile because `Property 'addCustomResourceDefinition' does not exist on type 'Api'`.

Installed versions:

- `kubernetes-client: 6.1.0`
- `typescript: 3.0.3`

And when I've looked into the definition code the `addCustomResourceDefinition` method is not present indeed:

```typescript
  interface Api {
    // Path templating
    // Sub-paths
    'apis': Apis
    'api': Apis
    'logs': Logs
    'log': Logs
    'version': Version
    // Calls
    loadSpec(): Promise<Api>
  }
```

jirihybek:
- Fix #334 - Add method `addCustomResourceDefinition` to the typings.
- Add `.../status` and `.../scale` endpoints to CRD swagger definition as specified in the Kubernetes API

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -5971,6 +5971,7 @@ declare namespace KubernetesClient {
     'version': Version
     // Calls
     loadSpec(): Promise<Api>
+    addCustomResourceDefinition(schema: object): void;
   }
 
   interface ApiClient {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -5971,7 +5971,7 @@ declare namespace KubernetesClient {
     'version': Version
     // Calls
     loadSpec(): Promise<Api>
-    addCustomResourceDefinition(schema: object): void;
+    addCustomResourceDefinition(schema: object): void
   }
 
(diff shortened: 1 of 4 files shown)


typings/index.d.ts
```diff
@@ -5971,6 +5971,7 @@ declare namespace KubernetesClient {
     'version': Version
     // Calls
     loadSpec(): Promise<Api>
+    addCustomResourceDefinition(schema: object): void;
```
silasbw: index.d.ts is machine generated.  Can you add this to the template (https://github.com/godaddy/kubernetes-client/blob/master/scripts/templates/ts-interface.mustache#L14)?

Issue #328, opened 2018-09-11
Mechanism to refresh token


baldred:
I'm wondering if there is any way we can use this kubernetes-client module to refresh authentication token when **auth-provider** is **oidc**. Below is the snippet of my kubeconfig (it's testing environment so it's okay to share in plaintext).

```
auth-provider:
      name: oidc
      config:
        client-id: bx
        client-secret: bx
        id-token: eyJraWQiOiIyMDE3MTAzMC0wMDowMDowMCIsImFsZyI6IlJTMjU2In0.eyJpYW1faWQiOiJJQk1pZC0zMTAwMDIxMlcxIiwiaXNzIjoiaHR0cHM6Ly9pYW0ubmcuYmx1ZW1peC5uZXQva3ViZXJuZXRlcyIsInN1YiI6ImJhbGRyZWRAaWQuaWJtLmNvbSIsImF1ZCI6ImJ4IiwiZ2l2ZW5fbmFtZSI6IkFsZHJlZCIsImZhbWlseV9uYW1lIjoiQmVuZWRpY3QiLCJuYW1lIjoiQWxkcmVkIEJlbmVkaWN0IiwiZW1haWwiOiJiYWxkcmVkQGlkLmlibS5jb20iLCJleHAiOjE1MzY1NzExMzEsInNjb3BlIjoiaWJtIG9wZW5pZCIsImlhdCI6MTUzNjU2NzUzMX0.S08p2PqmtLezWkPDJX7K-P54QMRmVKooMAKkL8gf4PL5NkLBV-zhumt3xmHzU2nN34Jkw9hlnALiEdDWY5ZC5a-ObXDy-nJpE65InvdF-icXQZknVOjqiKrrTDI5QR_a67vSrO1dgmDmjQ0Al7Nw9YxIqZPU5gKq5CSZcqqW_X9QYQl5sp5Qaec3wgY4ALOnnzuqQ38TwmHC_4P3QU0HF10Tz7f2nwYxSMyxwX8f5WGcanymZ8Mf7oSiCUMPEjUY9rnxTxEzQA-2Sf_Dv80dcQWZCXXwE4tgVbkUls98B3xDVaCsgs6WhplC87P1BaSrEMmHrn2nDfZKUwANcCwaWg
        idp-issuer-url: https://iam.ng.bluemix.net/kubernetes
[...]

PR #338, opened 2018-10-19
fix(auth): update auth settings after refreshing

silasbw:
cc https://github.com/godaddy/kubernetes-client/issues/328


silasbw:
Thanks @baldred

PR #344, opened 2018-10-21
querystring indices must be disabled

Maxpain177:


silasbw:
Hey, thanks for sending this :+1:

PR #356, opened 2018-11-04
Change the endpoint called by loadSpec

### LINKED ISSUE #352: Replace loadSpec URL to new OpenAPI spec
According to this link, the old swagger endpoints are deprecated as of Kubernetes 1.10:
https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-and-swagger-definitions

`loadSpec` should call `/openapi/v2` with appropriate headers.

It may be a good idea to implement this as using `/openapi/v2` primarily, and to fallback to the old URL in case the cluster has an old version of Kubernetes.

I can probably knock this out this weekend.

ajpauwels:
According to this Kubernetes doc:
https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-and-swagger-definitions

The `/swagger*` endpoints have been deprecated starting with v1.10.

This commit changes `loadSpec` to address this:
1. Initially attempts to contact `/openapi/v2` with `Accept: application/json` in order
   to get the swagger file via the new URL
2. If that fails with a 404, attempts to contact `swagger.json` to support clusters which
   do not have the new endpoint

fixes #352

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -37,18 +37,28 @@ class Root extends Component {
    * Load swagger.json from kube-apiserver specified via this.http.
    * @returns {Promise} Promise
    */
-  loadSpec() {
-    return new Promise((resolve, reject) => {
-      this.http.request('GET', { path: '/swagger.json' }, (err, res) => {
-        if (err) return reject(err);
-        if (res.statusCode !== 200) {
-          return reject(new Error(`Failed to get swagger.json: ${res.statusCode}`));
-        }
-        this._addSpec(res.body);
-        resolve(this);
-      });
-    });
-  }
+	loadSpec() {
+		return new Promise((resolve, reject) => {
+			this.http.request('GET', { headers: { 'accept': 'application/json' }, path: '/openapi/v2' }, (err, res) => {
+				if (err) return reject(err);
+				if (res.statusCode === 404) {
+					this.http.request('GET', { path: '/swagger.json' }, (err, res) => {
+						if (err) return reject(err);
+						if (res.statusCode !== 200) {
+							return reject(new Error(`Failed to get swagger.json: ${res.statusCode}`));
+						}
+						this._addSpec(res.body);
+						resolve(this);
+					});
+				}
+				else if (res.statusCode !== 200) {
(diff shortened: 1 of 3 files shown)


lib/swagger-client.js
```diff
@@ -39,10 +39,19 @@ class Root extends Component {
    */
   loadSpec() {
     return new Promise((resolve, reject) => {
-      this.http.request('GET', { path: '/swagger.json' }, (err, res) => {
+      this.http.request('GET', { headers: { accept: 'application/json' }, path: '/openapi/v2' }, (err, res) => {
```
silasbw: Is the `headers` required?  I would expect https://github.com/godaddy/kubernetes-client/blob/master/lib/request.js#L147 to do that implicitly.

lib/swagger-client.js
```diff
@@ -39,10 +39,19 @@ class Root extends Component {
...
         if (err) return reject(err);
-        if (res.statusCode !== 200) {
-          return reject(new Error(`Failed to get swagger.json: ${res.statusCode}`));
+        if (res.statusCode === 404) {
+          return this.http.request('GET', { path: '/swagger.json' }, (err, res) => {
```
silasbw: Can you run the linter?  I think it will complain about aliasing here (e.g., `err` in the first callback and `err` here).

lib/swagger-client.js
ajpauwels: You're right, it's unnecessary, I explicitly added it because the k8s docs (https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-and-swagger-definitions) use headers to differentiate between json and pb swaggers, but k8s also explicitly uses `application/json` if header is not set, will remove.

lib/swagger-client.js
ajpauwels: fixed `err` and `res` aliasing

ajpauwels:
What's your policy on `package-lock.json`, looks like mine changed when I ran `npm install`.

PS: Hope you had a good vacation!

silasbw:
Published this change with `kubernetes-client@6.4.0`.  Thanks again @ajpauwels :medal_sports:

PR #361, opened 2018-11-14
feat(config): add pathPrefix parameter to getInClusterMethod

NicolaiSchmid:
When using telepresence.io all mounted volumes are prefixed by a dynamic path. So to use this client with telepresence and the automatic configuration loading
the loading path has to be prefixed with the current TELEPRESENCE_ROOT env var. This change accounts for the problem and lets the user define a custom path prefix


silasbw:
This is old, contains a release, and changes the name of the package.  Is path prefix work is still valuable?

PR #364, opened 2018-11-22
Fallback to old swagger.json endpoint

### LINKED ISSUE #363: {url}/openapi/v2  receive a 403 error
Hi,

The last version of this library doesn't work with my cluster definition ( Openshift 3.7).

When We try to get the Api definition from {url}/openapi/v2  we receive a 403 error.

I saw this commit:

https://github.com/godaddy/kubernetes-client/commit/8a5f5cb46408f87bdd1b5687efb5b7e52b9df9b5

Could it be possible to support more code error in addition to 404 ?

Thanks in advance

ajpauwels:
Previously, only a 404 on /openapi/v2 would cause a fallback
to the /swagger.json endpoint. Now, ALL error codes from /openapi/v2
cause a fallback to swagger.json

fixes #363

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -40,17 +40,19 @@ class Root extends Component {
     return new Promise((resolve, reject) => {
       return this.http.request('GET', { path: '/openapi/v2' }, (modernGETErr, modernGETRes) => {
         if (modernGETErr) return reject(modernGETErr)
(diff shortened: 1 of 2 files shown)


silasbw:
Published with `kubernetes-client@6.4.1` -- thanks @ajpauwels

PR #368, opened 2018-12-10
refactor: remove spread for node6+ support

mushroom2:


lib/auth-providers/cmd.js
```diff
@@ -33,10 +33,7 @@ module.exports = {
       const cmdEnv = config['cmd-env']
 
       const output = spawnSync(cmd, args, {
-        env: {
-          ...process.env,
-          ...cmdEnv
-        },
+        env: Object.assign(process.env, cmdEnv),
```
silasbw: Can you change this to:

```js
Object.assign({}, process.env, cmdEnv),
```

so we avoid mutating `process.env`?

mushroom2: Sure. Done with it!

silasbw:
Published with `kubernetes-client@6.5.1`

PR #372, opened 2018-12-13
Remove authentication from /openapi/v2 call

### LINKED ISSUE #371: /openapi/v2 returns 401 when an authenticated call is made
If the `/openapi/v2` endpoint is called using a bearer token, then that token is checked and a 401 Unauthorized is returned if the token is invalid. This is poor behavior on K8s's part, but until then, the request to `/openapi/v2` needs to be made without a bearer token.

ajpauwels:
fixes #371

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -33,11 +33,7 @@ const Request = require('./request')
 
 class Root extends Component {
   _getSpec (pathname) {
-    const noAuthHTTP = new Request({
-      url: this.http.requestOptions.baseUrl,
-      ca: this.http.requestOptions.ca
-    })
-    return noAuthHTTP.http({ method: 'GET', pathname })
+    return this.http.http({ method: 'GET', pathname, noAuth: true })
       .then(res => {
         return res.body
       })
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -38,10 +38,14 @@ class Root extends Component {
    */
   loadSpec () {
     return new Promise((resolve, reject) => {
-      return this.http.request('GET', { path: '/openapi/v2' }, (modernGETErr, modernGETRes) => {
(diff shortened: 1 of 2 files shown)


lib/swagger-client.js
```diff
@@ -33,14 +33,14 @@ const Request = require('./request')
 
 class Root extends Component {
   _getSpec (pathname) {
-    return this.http.http({ method: 'GET', pathname })
+    const noAuthHTTP = new Request({
+      url: this.http.requestOptions.baseUrl,
+      ca: this.http.requestOptions.ca
+    })
+    return noAuthHTTP.http({ method: 'GET', pathname })
```
silasbw: What if we added a `noAuth` option to `http`?  One benefit of doing that is if we modify `Request` or add a new version and a user constructs it with some custom configuration that we wouldn't know how to replicate when we construct it here.

Thoughts?

lib/swagger-client.js
```diff
@@ -33,14 +33,14 @@ const Request = require('./request')
 
 class Root extends Component {
   _getSpec (pathname) {
-    return this.http.http({ method: 'GET', pathname })
```
silasbw: I was imagining something like:

```js
return this.http.http({ method: 'GET', pathname, noAuth: true })
```

lib/swagger-client.js
silasbw: Thoughts on https://github.com/godaddy/kubernetes-client/pull/372/files#r242026927 ?

ajpauwels:
A little swamped this week but will get to this over the weekend.

ajpauwels:
@silasbw Addressed comments in latest PR

silasbw:
Thanks again.  Published with `kubernetes-client@6.6.0`.

Issue #382, opened 2018-12-27
Support relative root when running client inCluster


olegsu:
Sometimes, when running API client from in-cluster config, the token-namespace-caPath are not injected into the default path (`/var/run/secrets/kubernetes.io/serviceaccount/...`).

For instance, I use [telepresence](https://telepresence.io) to develop locally.
Telepresence injects the file system of a pod into a `/tmp/` and exposes it as `TELEPRESENCE_ROOT`.
More detailed info here: https://www.telepresence.io/howto/volumes

I suggest allowing pass some environment variable to support relative root path.

PR #383, opened 2018-12-27
Allow passing relative root  when running inCluster

olegsu:
related to #382

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -6,9 +6,10 @@ const path = require('path')
 const yaml = require('js-yaml')
 const merge = require('deepmerge')
 
-const caPath = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt'
-const tokenPath = '/var/run/secrets/kubernetes.io/serviceaccount/token'
-const namespacePath = '/var/run/secrets/kubernetes.io/serviceaccount/namespace'
+const root = process.env.KUBERNETES_CLIENT_SERVICEACCOUNT_ROOT || '/var/run/secrets/kubernetes.io/serviceaccount/'
+const caPath = path.join(root, 'ca.crt')
+const tokenPath = path.join(root, 'token')
+const namespacePath = path.join(root, 'namespace')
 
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -32,16 +32,16 @@ function defaultConfigPaths () {
 function getInCluster () {
   const host = process.env.KUBERNETES_SERVICE_HOST
   const port = process.env.KUBERNETES_SERVICE_PORT
-
+  const root = process.env.RELATIVE_ROOT || '/'
```
silasbw: Can we pick a kubernetes-client specific name for this environment variable?  How about `KUBERNETES_CLIENT_SERVICEACCOUNT_ROOT`?

silasbw: Could we also set the default to `/var/run/secrets/kubernetes.io/serviceaccount/` and update `caPath`, `tokenPath`, and `namespacePath` to be like `${ KUBERNETES_CLIENT_SERVICEACCOUNT_ROOT }/*`?  I'm guessing that would implement the behavior folks have in mind when they set the `KUBERNETES_CLIENT_SERVICEACCOUNT_ROOT` value.

olegsu: Sounds good, will do asap

olegsu: Not sure I fully understand, can you please give an example?

olegsu:
@silasbw please check this out.
I've updated the name to be `KUBERNETES_CLIENT_SERVICEACCOUNT_ROOT`.
So it supports my example as:
```sh
export KUBERNETES_CLIENT_SERVICEACCOUNT_ROOT=$TELEPRESENCE_ROOT/var/run/secrets/kubernetes.io/serviceaccount/
```

lib/config.js
silasbw: The updates you made were what I had in mind. :+1:

silasbw:
Published your change with `kubernetes-client@6.7.0`.

olegsu:
Awesome, thanks

Issue #391, opened 2019-01-02
Latest version of swagger-fluent has verbose errors


weikinhuang:
I just updated `kubernetes-client` to latest and it brought in a newer version of swagger-client.

The client is printing out in my logs every time I make a request to the kubernetes api:

```
Component / already has a Path Item Object
```

Looks like swagger-fluent was updated 5 days ago:

https://github.com/silasbw/swagger-fluent/blob/master/lib/loader.js#L204

silasbw:
Thanks @weikinhuang -- I'm looking at this right now.

silasbw:
Published fix with `kubernetes-client@6.7.1`

PR #393, opened 2019-01-04
fix(aliasing): don't alias apis (to api)

silasbw:
/api is an endpoint, so we shouldn't alias /apis to /api.

Fixes https://github.com/godaddy/kubernetes-client/issues/391

### DIFF
--- a/lib/alias.js
+++ b/lib/alias.js
@@ -45,6 +45,10 @@ const esPlurals = {
   ingresses: true
 }
 
+const excludeFromAliasing = {
+  apis: true
+}
+
 module.exports = function (resourceType) {
   let aliases = [resourceType]
   if (resourceAliases[resourceType]) {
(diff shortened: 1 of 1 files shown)


hlian:
@silasbw This might need a slightly bigger callout in the CHANGELOG / or a deprecation notice with console.warn. I recently ran into this issue:

- Docker for Mac _has_ a `/apis/v1` implemented in its proprietary version of Kubernetes
- Official Kubernetes does not
- We upgraded to kubernetes-client 6.8. It worked hunky-dory against Docker for Mac
- It crashed when we pushed out to AWS, where it talked to a real Kubernetes instance
- I made a quick trip to the CHANGELOG
- Ended up here

Anyway, thanks for the library, it's been very good to us otherwise. :)

silasbw:
Oh no!  I'm sorry about that and you're right: I should have deprecated/warned this and made this a breaking change / major version bump.  Thanks for the feedback and I'll adjust my process for similar changes going forward.

For this specific issue, I can try to get something in place to prevent others from hitting the gotcha you ran into.  Thanks again.

PR #394, opened 2019-01-06
feat(backends): @kubernetes/client-node

silasbw:
Experimental support for a @kubernetes/client-node based backend.


silasbw:
FYI @brendandburns

PR #395, opened 2019-01-07
Adding token based authorization

jazzyarchitects:
If a developer wants to use a token associated with a service account as an authentication mechanism for kubernetes APIs then pass token as:

```js
const Client = require('kubernetes-client').Client
const client = new Client({
  config: {
    url: 'CLUSTER_URL',
    auth: {
      token: 'SERVICE ACCOUNT JWT'
    },
    insecureSkipTlsVerify: true,
  }
})
```

### DIFF
--- a/lib/backends/request.js
+++ b/lib/backends/request.js
@@ -177,6 +177,10 @@ class Request {
 
     if (options.noAuth) {
       delete requestOptions.auth
+    } else if (this.requestOptions.auth && this.requestOptions.auth.token) {
+      requestOptions.headers = requestOptions.headers || {}
+      requestOptions.headers.Authorization = `Bearer ${this.requestOptions.auth.token}`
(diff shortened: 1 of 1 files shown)


silasbw:
Thanks @jazzyarchitects.  Did you try something like this?  (Sorry this isn't documented well)

```js
const client = new Client({
  config: {
    url: 'CLUSTER_URL',
    auth: {
      bearer: 'bearerToken'
    },
    insecureSkipTlsVerify: true
  }
})
```

jazzyarchitects:
Thanks @silasbw. This works for me.

PriyankaTwinkel003:
hi @jazzyarchitects ,

i am new to k8s. i am working with GKE. How can i find the cluster URL and bearer token. An external client would call the app hosted on GKE.
so i have to share the details with the external service.

PR #396, opened 2019-01-09
swagger-client: Remove noAuth

whs:
The /openapi/v2 endpoint is protected by authentication. Also reported by one of my downstream user on https://github.com/wongnai/kube-slack/issues/50

I'm not sure why this call was flagged with noAuth?

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -33,7 +33,7 @@ const Request = require('./backends/request')
 
 class Root extends Component {
   _getSpec (pathname) {
-    return this.backend.http({ method: 'GET', pathname, noAuth: true })
+    return this.backend.http({ method: 'GET', pathname })
       .then(res => {
         return res.body
       })


silasbw:
@ajpauwels can you help reconcile this with https://github.com/godaddy/kubernetes-client/pull/372 ?

whs:
I tested /openapi/v2 and it does not requires authentication in GKE. However, in my cluster that use Kops it does require authentication. It might be due to different API server flag.

From what I see the reason authentication was removed was due to #371 as the user pass invalid credential to the library and get a server error. I think this is entirely user's error as it is a [documented](https://v1-12.docs.kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-requests) behavior.

silasbw:
Thanks @whs :100: .

Published your fix with `kubernetes-client@6.8.1`.

PR #397, opened 2019-01-11
fix(require): update auth-provider path

silasbw:


silasbw:
cc @gordlea

Thanks for reporting :)

silasbw:
Published with `kubernetes-client@6.8.0` -- can you confirm it fixes the issue you're seeing?

PR #415, opened 2019-02-21
Add swagger files for latest releases

pastudan:
For future reference, these files are here: https://github.com/kubernetes/kubernetes/blob/release-1.13/api/openapi-spec/swagger.json

I just downloaded and gzipped them for this PR


silasbw:
Published with `kubernetes-client@6.9.0`, thanks again!

PR #420, opened 2019-03-01
feat: ability for a user to define aliases

lholmquist:
This allows a user to specify their own aliases when creating a Client

The motivation for this comes from the [openshift node client ](https://github.com/nodeshift/openshift-rest-client)that my team has.  We are refactoring it to have it generate the client automatically by passing in the swagger spec.  https://github.com/nodeshift/openshift-rest-client/pull/113

There are some endpoints that openshift has, that if aliased, would be easier to work with

For example,  there is an endpoint `build.openshift.io`, which when generated into an api would be some like this: `client.api['build.openshift.io'].v1.get()`

If `build.openshift.io` was aliased to just be `build` then it would make the client api easier to use for the user, since it could then just be `client.api.build.v1.get()`

I need to add something to the readme to show how to do this, but wanted to get some thoughts on this first.

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -28,7 +28,7 @@ const fs = require('fs')
 const path = require('path')
 const zlib = require('zlib')
 
-const getAliases = require('./alias')
+const alias = require('./alias')
 const Request = require('./backends/request')
 
 class Root extends Component {
@@ -138,6 +138,8 @@ class Client {
         `swagger-${options.version}.json.gz`)
       spec = JSON.parse(zlib.gunzipSync(fs.readFileSync(swaggerPath)))
     }
+
+    const getAliases = alias(options.aliases)
 
     const root = new Root({ splits: [], backend, getNames: getAliases })
     if (spec) root._addSpec(spec)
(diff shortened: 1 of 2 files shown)


lib/swagger-client.js
```diff
@@ -139,6 +139,8 @@ class Client {
       spec = JSON.parse(zlib.gunzipSync(fs.readFileSync(swaggerPath)))
     }
 
+    const getAliases = alias(options.aliases)
```
silasbw: What if you write your own `getNames` function in https://github.com/nodeshift/openshift-rest-client and we update the `Client` constructor to allow developers to pass in a customer `getNames` function?  That might look something like:

```js
const kubernetesAliases = require('kubernetes-client/alias')
const aliases = {
  'build.openshift.io': ['build']
};

function openshiftAliases(name) {
  if (name in aliases) return aliases.concat([name])
  return kubernetesAliases(name)
}

class OpenshiftClient extends Client {
  constructor(options) {
    return super(Object.assign({ getNames: openshiftAliases }, options))
  }
}
```

lholmquist: @silasbw I guess that does look a bit cleaner, if we were able to pass in a `getNames` function.

Do you want me to modify this PR, or should i open a new issue/PR since it is a different thing?

silasbw:
We're going with https://github.com/godaddy/kubernetes-client/pull/424.

Issue #418, opened 2019-02-21
User Configurable Aliases


lholmquist:
It didn't look like there was a way currently to pass an object to add more aliases.

Context:  I created a [rest client for Openshift](https://github.com/nodeshift/openshift-rest-client), and i'm trying to make the code simpler.  I've actually passed this library a openshift swagger doc, and things just worked!  

Openshift has its own aliases for things, like `bc` for `buildconfigs` and `is` for `ImageStreams`.

Just taking a quick glance at the alias code, it doesn't look like it would be much work to pass something in and merge what is currently defined.  

I'd be happy to submit a PR for it

lholmquist:
closing.  part of #424

PR #424, opened 2019-03-04
feat:  ability for a user to specify their own getNames funciton.

lholmquist:
getNames is used to setup alias's for the client.  This would allow someone to specify their own aliases.  This relates to #418

This PR should replace #420 

The changes here are much more minimal than the previous changes.

@silasbw let me know what you think


silasbw:
published with kubernetes-client@6.10.0 -- thanks again @lholmquist!

lholmquist:
@silasbw cool. Thanks!

PR #427, opened 2019-03-07
feat: have the ability to make the request json option false

lholmquist:
One of the endpoints we have in openshift, "initiatebinary", allows someone to upload a tar archive to openshift as part of there build.  In order to make it work with request, you have to specify the json option to be false.

Currently the default options in the `http` method have the json value set to true with no way to change it.  

This PR adds the ability to change that value to false if needed.

### DIFF
--- a/lib/backends/request.js
+++ b/lib/backends/request.js
@@ -171,7 +171,7 @@ class Request {
       method: options.method,
       uri,
       body: options.body,
-      json: true,
+      json: 'json' in options ? Boolean(options.json) : true,
       qs: options.parameters || options.qs,
       headers: options.headers
     }, this.requestOptions)


lib/backends/request.js
```diff
@@ -171,7 +171,7 @@ class Request {
       method: options.method,
       uri,
       body: options.body,
-      json: true,
+      json: !(options.json === false),
```
silasbw: Do you want any falsy value for `options.json` to disable json (e.g., `undefined`), or only `false`?  If it's the former, you can make this change:

```js
const json = 'json' in options ? Boolean(options.json) : true
```

I'm good with either approach.

lholmquist: it should default to true,  so undefined in this case would also be true.


If you want me to change it to what you suggested let me know

silasbw: In the code i posted `{ json: undefined }` results in setting json to `false`.  `{ }` (i.e., no `.json` in options) results in setting json to `true`.

lholmquist: yea,  i guess i meant, if no `.json` is passed in, it should be true.  if you were passing in undefined, which i hope someone doesn't do, it should be false

lholmquist: I've updated and squashed the other commits

silasbw:
Published with kubernetes-client@6.11.0

PR #428, opened 2019-03-09
Add vpa specific example of crd usage

naseemkullah:
cc @silasbw


silasbw:
Thanks @naseemkullah :clinking_glasses:

naseemkullah:
My pleasure!

PR #431, opened 2019-03-22
Updated js-yaml to 3.13.0 to patch npm vulnerability in advisory 788

charlenetshos:
npm has reported denial of service vulnerability that's affecting versions of `js-yaml` before 3.13.0.  This caused `npm audit` to fail on our modules that rely on `kubernetes-client`. 

There is an explanation of the vulnerability on this advisory: https://www.npmjs.com/advisories/788
This PR bumps the version of `js-yaml` as per the advisory's recommendation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3406,9 +3406,9 @@
       "dev": true
     },
     "js-yaml": {
-      "version": "3.12.0",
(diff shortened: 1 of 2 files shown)


silasbw:
Published with kubernetes-client@6.11.1.  Thanks!

PR #470, opened 2019-05-19
fix(CRDs): malformed watch paths for namespaced CRDs

### LINKED ISSUE #468: watching custom resources with a namespace scope
Using the deployment-notifier.js example works but if I change
```
const stream = client.apis['kubernetes-client.io'].v1.watch.deploymentnotifiers.getStream()
```
to
```
const stream = client.apis['kubernetes-client.io'].v1.namespaces('default').watch.deploymentnotifiers.getStream()
```
the script returns immediately without any output.

Trying to do the same thing in another project with a CRD named *toolchain* which has a service account with a role which only allows access to that resource I get the following response.
```
{
  "kind": "Status",
   "apiVersion": "v1",
   "metadata": {},
   "status": "Failure",
   "message": "watch.stable.liatr.io \"toolchains\" is forbidden: User \"system:serviceaccount:toolchain:operator-slack\" cannot get resource \"watch\" in API group \"stable.liatr.io\" in the namespace \"toolchain\"",
   "reason": "Forbidden",
   "details": {
     "name": "toolchains",
     "group": "stable.liatr.io",
     "kind": "watch"
   },
   "code": 403
 }
```
If I add a resource named *watch* to my role I do not get any response at all so my guess is that the client is sending "watch" instead of the resource name to the Kubernetes API.

iffyio:
Currently, URLs for namespaced watches for CRD
objects are incorrectly formatted (the result instead
tries to do a GET on a CRD named `watch`), causing k8s to
return a 404.
This patch corrects this issue

Fixes #468

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -93,8 +93,8 @@ class Root extends Component {
 
     const watchPaths = {
       watchCluster: `/apis/${group}/${version}/watch/${name}`,
-      watchNamespace: `/apis/${group}/${version}${namespace}/watch/${name}`,
-      watchResource: `/apis/${group}/${version}${namespace}/watch/${name}/{name}`
+      watchNamespace: `/apis/${group}/${version}/watch${namespace}/${name}`,
+      watchResource: `/apis/${group}/${version}/watch${namespace}/${name}/{name}`
     }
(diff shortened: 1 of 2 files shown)


silasbw:
`kubernetes-client@8.0.0` includes this fix.

PR #484, opened 2019-05-25
fix(CRDs): Use `versions` list for spec paths

### LINKED ISSUE #458: How do I get Custom Resource Definition properties?
I created a namespaced scope CRD. 
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: payloads.hook-to-k8s.sfxworks.net
spec:
  group: hook-to-k8s.sfxworks.net
  versions:
    - name: v1
      served: true
      storage: true
  scope: Namespaced
  names:
    plural: payloads
    singular: payload
    kind: Payload
    shortNames:
    - pl
    - payl
```

I am able to run `kubectl-proxy` and see that my CRD is active via http://127.0.0.1:8001/apis/hook-to-k8s.sfxworks.net/v1

With the client configured to use a service account in the pod and the pod configured to use a service account that has the proper Role/Role bindings, I get an error when trying to access this resource.

Using
`const payloads = await client.apis["hook-to-k8s"].sfxworks.net.v1.namespaces.default.payloads.get()`
I get 
```
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'sfxworks' of undefined
```
I have tried renaming this several ways, including
`  const payloads = await client.apis["hook-to-k8s.sfxworks.net"].v1.namespaces.default.payloads.get() `

and I still get undefined when trying to access any object after apis. How do I retrieve and post to my resource?

iffyio:
Currently, it's only possible to specify one api
version for CRD objects via the deprecated
`version` field when building their spec paths.
This commit adds support for building spec paths
for multiple CRD versions specified in the
`versions` list while maintaining compatibility
with `version`.

Fixes #458

### DIFF
--- a/examples/crontabs-crd.json
+++ b/examples/crontabs-crd.json
@@ -2,8 +2,12 @@
   "kind": "CustomResourceDefinition", 
   "spec": {
     "scope": "Namespaced", 
-    "version": "v1", 
-    "group": "stable.example.com", 
+    "versions": [{
+      "name": "v1",
+      "served": true,
+      "storage": true
+    }],
+    "group": "stable.example.com",
     "names": {
       "shortNames": [
         "ct"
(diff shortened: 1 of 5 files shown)


silasbw:
Published with `kubernetes-client@8.1.3`

PR #492, opened 2019-06-07
test(patch): add integration test

celiawaggoner:


test-integration/patch.js
```diff
@@ -0,0 +1,43 @@
...
+    let pod
+    const manifest = yaml.safeLoad(fs.readFileSync('./test-integration/busybox-pod.yaml'))
+    const name = manifest.metadata.name
+    const client = await env.getClient()
+    await client.api.v1.namespaces(namespace).pods.post({ body: manifest })
+
+    pod = await client.api.v1.namespaces(namespace).pods(name).get()
+    expect(pod.body.spec.containers[0].image).equals('busybox')
+
+    await client.api.v1.namespaces(namespace).pods(name).patch({
```
silasbw: nit: use a different patch body. this one could cause a docker to pull a new image to the test machine, or maybe we're already using `busybox:musl` (and the patch could be failing). patching annotations is a common thing to do in applications that use this library:

```js
await client.api.v1.namespaces(namespace).pods(name).patch({
  body: {
    metadata: { 
      annotations: {
        'patch-test0': process.pid
      }
    }
  } 
})
```

test-integration/patch.js
```diff
@@ -0,0 +1,43 @@
...
+          containers: [{
+            name: 'busybox',
+            image: 'busybox:musl'
+          }]
+        }
+      }
+    })
+
+    pod = await client.api.v1.namespaces(namespace).pods(name).get()
+    expect(pod.body.spec.containers[0].image).equals('busybox:musl')
```
silasbw: Update this too.

PR #507, opened 2019-06-25
Add shell option to spawnSync for Windows Support

jeremymgibbons:
Specs:
OS: Windows 10
Shell: Git Bash
Was receiving this error message. 
{ error:
   { Error: spawnSync gcloud ENOENT
    at _errnoException (util.js:1022:11)
    at spawnSync (child_process.js:579:20)
    at Promise (C:\dev\brandify-tools\node_modules\kubernetes-client\lib\auth-providers\cmd.js:38:22)
    at new Promise (<anonymous>)
    at Object.refresh (C:\dev\brandify-tools\node_modules\kubernetes-client\lib\auth-providers\cmd.js:30:12)
    at Promise (C:\dev\brandify-tools\node_modules\kubernetes-client\backends\request\client.js:21:14)
    at new Promise (<anonymous>)
    at refreshAuth (C:\dev\brandify-tools\node_modules\kubernetes-client\backends\request\client.js:19:10)
    at Request.request [as _callback] (C:\dev\brandify-tools\node_modules\kubernetes-client\backends\request\client.js:150:16)
    at Request.self.callback (C:\dev\brandify-tools\node_modules\request\request.js:185:22)
     code: 'ENOENT',
     errno: 'ENOENT',
     syscall: 'spawnSync gcloud',
     path: 'gcloud',
     spawnargs: [ 'config', 'config-helper', '--format=json' ] }

The pull request fixes this issue for spawnSync on Windows.


lib/auth-providers/cmd.js
```diff
@@ -34,7 +34,8 @@ module.exports = {
 
       const output = spawnSync(cmd, args, {
         env: Object.assign({}, process.env, cmdEnv),
-        windowsHide: true
+        windowsHide: true,
+        shell: true
```
jeffpearce: Can we only do this for Windows machines? I'd hate to change this for other platforms if not needed

jeremymgibbons:
It works for Mac with or without the shell options.

jeffpearce:
My concern is around changing the behavior, rather than correctness - if we don't need to run the command in a shell on mac and linux, I'd rather not. It's probably fine, but since it's already been well tested on those platforms I'd prefer not to change them there.

jeremymgibbons:
I see what you mean. Have you heard of any other windows users having issues? This is the only solution I have found for it to work on my machine. Any suggestions?

One solution I could try is checking the OS and passing in the shell options only if the platform is windows. And thoughts?

PR #547, opened 2019-10-18
Docs: Typescript Example Fix

francardoso93:
`import * as Api from 'kubernetes-client';
`
will print a warning in console for newer versions: 

> deprecated require('kubernetes-client').config, use require('kubernetes-client/backends/request').config

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,11 +144,10 @@ kubernetes-client includes a typings declartion file for Kubernetes
 API 1.13 and a complimentry `Client1_13` class:
 
 ```typescript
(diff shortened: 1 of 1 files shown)


lihaiswu:
@francardoso93 @silasbw  I got error `kubernetes-client/typings/index"' has no default export` while using `import ApiClient from 'kubernetes-client';`

lihaiswu:
Need to keep using `import * as Api from 'kubernetes-client';` and just change `Api.Client` to `Api.Client1_13`

PR #559, opened 2019-11-05
Throw errors if pump sees one

zoidbergwill:


zoidbergwill:
Any chance of getting these merged?

PR #571, opened 2020-01-06
Define rejectUnauthorized for WebSockets

kassner:
I'm fairly new to NodeJS ecossystem, and I'm trying to update a GKE cluster using Google Functions. Most of it works, but now that I'm trying to upgrade to `kubernetes-client:8.3.6`, I'm getting some TLS errors on the pod exec API:

```js
await client.api.v1.namespaces('www').pods.get().then(async podList => {
    await podList.body.items.forEach(async pod => {
        console.log(pod.metadata.name);
    })
})

const response = await client.api.v1.namespaces('www').pods('varnish-75fc8fbdf5-vkjxz').exec.post({
    qs: {
        command: ['id'],
        container: 'varnish',
        stdout: true,
        stderr: true
    },
    headers: {
        authorization: `Bearer ${client.backend.requestOptions.auth.bearer}`
    }
});

await response.messages.forEach(async message => {
    if (message.channel === 'stdout') {
        console.log(message.message);
    }

    if (message.channel === 'stderr') {
        console.error(message.message);
    }
})
```

With this code, I get the following:

```
varnish-75fc8fbdf5-vkjxz
{ Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1058:34)
    at TLSSocket.emit (events.js:198:13)
    at TLSSocket._finishInit (_tls_wrap.js:636:8) code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE', messages: [] }
```

Given `pods(name).exec.post()` runs within a WebSocket connection, apparently `strictSSL` is not enough for it, and `rejectUnauthorized` is necessary.

With the PR's change, now I get the following result, as I expected:

```
varnish-75fc8fbdf5-vkjxz

uid=1000(www-data) gid=1000(www-data) groups=1000(www-data)

```

Setting the env `NODE_TLS_REJECT_UNAUTHORIZED=1` also works, but then it applies to all TLS connections, not just the ones to GKE, which in my case it's a bad idea, because the function connects to another API.

PS: I've read the CONTRIBUTING.md, but I have no idea how to test this thing. I can try it out with some guidance.


kassner:
@silasbw how can I write tests for this?

PR #590, opened 2020-03-20
Update dependency

inductor:
acorn has moderate level vulnerability


silasbw:
ty!

I think https://github.com/godaddy/kubernetes-client/pull/593 addresses this.

